PR #33, opened 2025-02-10
Change forwardauth endpoint in example compose

robotman4:
The current endpoint in the example compose file does not work. All documentation points to /api/auth too, so I guess it should not be /api/auth/traefik.


steveiliop56:
Oh yes you are correct, it is there for v3.0.0 which will be like this but until then it is indeed incorrect. Can you also update the dev docker compose or should I?

robotman4:
Added dev compose too. Thanks for tinyauth :)

steveiliop56:
Thank you for the pr!

steveiliop56 (approved):

PR #44, opened 2025-03-04
feat: add Remote-User header

WilliamB78:
This PR adds a really simple Remote-User header feature.

Allowing traefik to formard the Remote-User header, the service proxied will be able to read from it and potentially authenticate the user inside the header.

I tested it with Zitadel as a Generic OAuth provider with a scope "email" and Firefly-III.


steveiliop56:
Thank you! Added some docs here too https://github.com/steveiliop56/tinyauth-docs/pull/4

PR #274, opened 2025-07-16
feat: better health check and less log noise

11notes:


steveiliop56:
Hello!

Thank you for the pull request. I can't help saying that I am a bit skeptical about it. I believe the best idea would be to instead log the healthcheck in debug mode instead of completely blacklisting it. Additionally I see you have included some code/comments that I previously removed. Was that on purpose or accidental?

11notes:
Huh? Which comments did I readd? I forked the latest branch:

image

But I see that there are lots of deltas. I try to resync.

As for the log, if you want to move them to debug that's an option too, I just don't think that a health check or favicon are relevant in any way.

steveiliop56:
The latest version of `server.go` is [here](https://github.com/steveiliop56/tinyauth/blob/main/internal/server/server.go). Your pull request reverts some of my changes like renaming `handlers.TOTPHandler` back to `handlers.TotpHandler`. As for the logging, I believe that the healthcheck in debug can be useful because you can confirm if the service is being reached. Can you use a different path? Sure but since the debug mode exists why not utilize it?

11notes:
Really odd that I got the wrong server.go when forking and cloning to my development VM, but so be it, I’ve updated it with the current file and moved the log for matches to debug. Now you have less noise when run in production but all the noise if using debug.

steveiliop56:
Perfect!

PR #381, opened 2025-10-03
Distroless dockerfile

dallemon:
this pr is a minor improvement to the current dockerfile

- use chainguard static distroless image for runner
- compress binary using upx to decrease final image size
- remove curl, leftover from old healthcheck
- align golang image in dev dockerfile
- pin all images to minor versions and specific sha256

## Summary by CodeRabbit

* **Chores**
  * Pinned container base images with digests for reproducible builds.
  * Updated builder and runtime images and standardized the entrypoint location.
  * Removed unnecessary tools to slim down the final image.
* **Refactor**
  * Streamlined the build workflow targeting Linux and a static runtime.
  * Adjusted binary paths and build steps for consistent packaging.
* **Impact**
  * Smaller container image and potentially faster startup.
  * More consistent, predictable deployments across environments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,8 +27,6 @@ ARG VERSION
 ARG COMMIT_HASH
 ARG BUILD_TIMESTAMP
 
-RUN apk add upx
-
 WORKDIR /tinyauth
 
 COPY go.mod ./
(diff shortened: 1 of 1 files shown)


steveiliop56:
Hello @dallemon,

While distroless is planned for a future release, this is not the approach I would like to follow. Making the main image distroless is a bad move because non-experienced users will have a hard time debugging since they will not know how to use `nsenter`. Ideally a new tag should be used for a distroless/rootless image (`v4-distoless`) but I haven't decided on a build workflow. Additionally building the app 6 times (arm64/amd64 binary + arm64/amd64 docker image + arm64/amd64 rootless) would take more time and waste resources. The approach I will probably take is to copy the binary from the previous steps to an empty distroless image and packaging it. That said, I will have to reject this pull request for now but it's definitely something I would like to revisit in the future.

PR #382, opened 2025-10-03
Update some workflows

dallemon:
this pr updates some of the workflows

- use digests for most github actions (will look at nightly and release later)
- add workflow for updating docker image digests
- simplified ci workflow a bit
- enabled dependabot updates for github actions

## Summary by CodeRabbit

- Chores
  - Pinned GitHub Actions to specific commit SHAs for improved reliability.
  - Added Dependabot configuration to auto-update GitHub Actions daily.
  - Introduced a scheduled workflow to automatically refresh Docker image digests.
  - Updated CI to Go 1.25 and streamlined frontend build to install, build, and copy assets in one step with a new asset path.
  - Aligned sponsors and stale workflows with pinned action versions and minor naming tweaks.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,3 +24,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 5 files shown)


steveiliop56:
Hello @dallemon,

Really appreciate the pull request, but I unfortunately cannot merge. I am not a big fan of using digests because they make the developer experience worse and they are harder to debug in case of a problematic version (I will have to look up the digests). Additionally I will have to depend both on an additional action (digest updater) and on a package of an external vendor (chainguard). While I understand the additional security digests provide, I believe that it's not worth sacrificing the developer experience for them.

PR #390, opened 2025-10-06
refactor: make cli modular

steveiliop56:


cmd/create.go
```diff
@@ -0,0 +1,99 @@
...
+		if err != nil {
+			log.Fatal().Err(err).Msg("Form failed")
+		}
+	}
+
+	if c.username == "" || c.password == "" {
+		log.Fatal().Err(errors.New("error invalid input")).Msg("Username and password cannot be empty")
+	}
+
+	log.Info().Str("username", c.username).Str("password", c.password).Bool("docker", c.docker).Msg("Creating user")
```
Rycochet: Does this want to actually log the password, or something like `"redacted"`? The hashed password is logged below which is good

cmd/root.go
```diff
@@ -15,55 +13,28 @@ import (
...
 }
 
-func init() {
-	rootCmd.AddCommand(userCmd.UserCmd())
-	rootCmd.AddCommand(totpCmd.TotpCmd())
+func (c *rootCmd) Register() {
+	c.cmd = &cobra.Command{
+		Use:   "tinyauth",
+		Short: "The simplest way to protect your apps with a login screen",
+		Long:  `Tinyauth is a simple authentication middleware that adds a simple login screen or OAuth with Google, Github and any provider to all of your docker apps.`,
```
Rycochet:
```suggestion
		Long:  `Tinyauth is a simple authentication middleware that adds a simple login screen or OAuth with Google, Github or any other provider to all of your docker apps.`,
```

cmd/verify.go
```diff
@@ -0,0 +1,126 @@
...
+						return errors.New("username cannot be empty")
+					}
+					return nil
+				})),
+				huh.NewInput().Title("Password").Value(&c.password).Validate((func(s string) error {
+					if s == "" {
+						return errors.New("password cannot be empty")
+					}
+					return nil
+				})),
```
steveiliop56, line 78: It's the user as in `user1:foo`, the username is the `user1` and the password is your password in plain text.

cmd/verify.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (c *verifyUserCmd) Register() {
+	c.cmd = &cobra.Command{
+		Use:   "verify",
+		Short: "Verify a user is set up correctly",
+		Long:  `Verify a user is set up correctly meaning that it has a correct username, password and TOTP code.`,
+		Run:   c.run,
+	}
+
+	c.cmd.Flags().BoolVarP(&c.interactive, "interactive", "i", false, "Create a user interactively")
```
Rycochet:
```suggestion
	c.cmd.Flags().BoolVarP(&c.interactive, "interactive", "i", false, "Validate a user interactively")
```

cmd/create.go
steveiliop56: Good catch.

cmd/root.go
```diff
@@ -15,55 +13,28 @@ import (
...
-		// Run
-		err = app.Setup()
-
-		if err != nil {
-			log.Fatal().Err(err).Msg("Failed to setup app")
-		}
-
-	},
+	viper *viper.Viper
 }
```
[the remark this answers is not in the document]
steveiliop56: We keep it so as all commands implement the same interface, could be used in the future.

PR #409, opened 2025-10-12
fix: allow root cookie domain host redirects

### LINKED ISSUE #408: [BUG] v4.0.0 - Inconsistencies with multi-level domain redirects
**Describe the bug**
After upgrade to v4, I've noticed inconsistent behaviour with redirections through Traefik forwardauth (kubernetes).

In my configuration, I have all my services stemming from a 3rd level domain, for example `cluster.domain.com`. Most of my services, including tinyauth are at the 4th level. E.g. `auth.cluster.domain.com` and `service1.cluster.domain.com`. 

The odd one out is the main landing page, which is served directly from the 3rd level (`cluster.domain.com`). All services at the 4th level work as expected. However, the 3rd level redirection does not work, instead of being redirected back to `cluster.domain.com` the browser stays on `auth.cluster.domain.com`.

**To Reproduce**
1. Configure traefik/tinyauth as described above.
2. Go to the 3rd level domain (unauthenticated)
3. Complete auth

**Expected behavior**
Should be redirected back to 3rd level.

**Device (please complete the following information):**
 - OS: Linux (NixOS)
 - Browser: Zen (Firefox)
 - Tinyauth: v4.0.0
 - Docker: N/A - Kubernetes v1.38.3

scottmckendry:
Fixes #408 

Previously IsRedirectSafe rejected redirects to the exact cookie domain when AppURL had multiple subdomain levels, because it stripped the first label twice.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved redirect validation: destinations that exactly match the expected host or derive from its cookie domain are now correctly allowed, fixing root-domain redirect handling and making subdomain decisions more consistent.

* **Tests**
  * Expanded test coverage for multi-level subdomains and malformed URLs to verify corrected redirect behavior and edge cases.

### DIFF
--- a/internal/utils/app_utils.go
+++ b/internal/utils/app_utils.go
@@ -100,17 +100,17 @@ func IsRedirectSafe(redirectURL string, domain string) bool {
 		return false
 	}
 
-	cookieDomain, err := GetCookieDomain(redirectURL)
+	host := parsedURL.Hostname()
+	if host == domain {
+		return true
+	}
 
+	cookieDomain, err := GetCookieDomain(redirectURL)
 	if err != nil {
 		return false
 	}
 
-	if cookieDomain != domain {
-		return false
-	}
-
-	return true
+	return cookieDomain == domain
 }
 
 func GetLogLevel(level string) zerolog.Level {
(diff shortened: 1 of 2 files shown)


scottmckendry:
@steveiliop56 since this is a change in behaviour for what appears to be some kind a security improvement introduced in v4, I've tried not to alter things _too_ much.

However, currently there is still an issue where if you have higher level domains, these will be rejected. E.g. `foo.bar.baz.example.com`. We could implicitly trust all domain levels off of the host, but I think I need to understand the security implications better before doing so.

steveiliop56:
Thank you!

PR #422, opened 2025-10-19
feat: ACL labels from environment variables

chrellrich:
## This pull request adds support for reading [ACL labels](https://tinyauth.app/docs/reference/labels) from environment variables.

As discussed in https://github.com/steveiliop56/tinyauth/issues/356 the ability to use the features that labels allow ( limiting access, allowing some paths without authentication, etc. ) for apps that are not running in docker would be great.

This is a first implementation of reading labels from environment variables and would solve that issue.

The syntax is the same as Docker labels but **uppercase and with dashes instead of dots**.
For example: `tinyauth.apps.[app].users.allow` becomes `TINYAUTH_APPS_[APP]_USERS_ALLOW`

You can use both env-labels and Docker labels simultaneously. If both exist for an app, env-labels are prioritized.

> [!WARNING]  
> Labels are not merged, so you can't set `tinyauth.apps.[app].users.allow` as a Docker label and `TINYAUTH_APPS_[APP]_IP_ALLOW` as an enviroment variable. In this case only the ip-allow config would be effective.

The feature is implemented as an additional service `LabelService` that is called by `proxy_controller`.

### Example for Docker compose
The access to whoami is controlled with a label set as a environment variable on the tinyauth container.

In this example only john is allowed access to whoami but not sam.

| Username | Password |
|--------|--------|
| sam | sam |
| john | john |

```
services:
  traefik:
    image: traefik:v3.3
    container_name: traefik
    command: --api.insecure=true --providers.docker
    restart: unless-stopped
    ports:
      - 80:80
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock

  whoami:
    image: traefik/whoami:latest
    container_name: whoami
    restart: unless-stopped
    labels:
      traefik.enable: true
      traefik.http.routers.whoami.rule: Host(`whoami.example.com`)
      traefik.http.routers.whoami.middlewares: tinyauth

  tinyauth:
    image: ghcr.io/chrellrich/tinyauth:label-from-env
    container_name: tinyauth
    restart: unless-stopped
    environment:
      - APP_URL=https://tinyauth.example.com
      - USERS=sam:$$2a$$10$$1GtofEBYXeSwKhOxEKxkOuEE6rK4ii5g3K/Q4hw03lR2NBwfffHae,john:$$2a$$10$$YSYms0GckSJ8oiuh1Cgmx.Wu1pkPu3LKdn0e3LRM7dp2k3jYFPaqi
      - TINYAUTH_APPS_WHOAMI_CONFIG_DOMAIN=whoami.example.com
      - TINYAUTH_APPS_WHOAMI_USERS_ALLOW=john
    labels:
      traefik.enable: true
      traefik.http.routers.tinyauth.rule: Host(`tinyauth.example.com`)
      traefik.http.middlewares.tinyauth.forwardauth.address: http://tinyauth:3000/api/auth/traefik
``` 

## Summary by CodeRabbit

* **New Features**
  * Access controls can now be configured via environment variables, providing an alternative to Docker labels for managing application access policies and permissions.


steveiliop56 (approved):

steveiliop56:
Thank you @chrellrich!

PR #431, opened 2025-10-25
refactor: rework decoders logic for cleaner code

steveiliop56:
## Summary by CodeRabbit

* **Refactor**
  * Internal configuration decoding was reworked to use generic, type-driven logic. No user-facing behavior or public API changes.

* **Tests**
  * Updated decoder tests and removed an older normalization test suite.

* **Chores**
  * Added an indirect dependency for string case utilities.


internal/utils/decoders/flags_decoder_test.go
```diff
@@ -9,52 +9,29 @@ import (
 )
 
 func TestDecodeFlags(t *testing.T) {
-	// Variables
+	// Setup
+	flags := map[string]string{
+		"--providers-google-client-id":        "google-client-id",
```
Rycochet, line 14: If cleaning up the tests making the values for these into shared constants rather than duplicating "google-client-id" here and line 23 etc can help make the tests a bit more obvious in what they're doing - identical result just nicer implementation

steveiliop56: Sure.

internal/utils/decoders/decoders.go
```diff
@@ -35,55 +32,76 @@ func NormalizeKeys(keys map[string]string, rootName string, sep string) map[stri
...
 			if i == 0 {
-				camelClientName += p
+				final += kebabToCamel(part)
 				continue
 			}
-			if p == "" {
-				continue
-			}
-			camelClientName += strings.ToUpper(string([]rune(p)[0])) + string([]rune(p)[1:])
+			final += "." + kebabToCamel(part)
```
Rycochet: This is appending strings in two lines with almost duplicate code, slightly cleaner might be -

```suggestion
			if i > 0 {
				final += "."
			}
			final += kebabToCamel(part)
```

PR #506, opened 2025-12-09
database: Show additional warning when sqlite file cannot be opened

plaes:
Currently, when database file cannot be created and opened, we get the following confusing error:
"failed to initialize database service: unable to open database file: out of memory (14)"

Unfortunately, the real error is either in Go's SQLite library or the ORM library as SQLite's errno=14 is `SQLITE_CANTOPEN`, based on https://sqlite.org/rescode.html#cantopen

For now, the best thing we can do is show a better though still somewhat confusing error message ourselves, until someone manages to figure out the real cause for this confusion.

## Release Notes

* **Chores**
  * Enhanced internal error logging for database connection issues to improve troubleshooting and diagnostic capabilities.

### DIFF
--- a/internal/service/database_service.go
+++ b/internal/service/database_service.go
@@ -4,6 +4,8 @@ import (
 	"database/sql"
 	"tinyauth/internal/assets"
 
+	"github.com/rs/zerolog/log"
+
 	"github.com/glebarez/sqlite"
 	"github.com/golang-migrate/migrate/v4"
 	sqliteMigrate "github.com/golang-migrate/migrate/v4/database/sqlite3"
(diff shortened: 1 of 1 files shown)


steveiliop56:
Thank you for the pull request! Not sure if that's needed though since a message will be [logged](https://github.com/steveiliop56/tinyauth/blob/main/internal%2Fbootstrap%2Fapp_bootstrap.go#L134-L136) in case the database initialization fails.

plaes:
> Thank you for the pull request! Not sure if that's needed though since a message will be [logged](https://github.com/steveiliop56/tinyauth/blob/main/internal%2Fbootstrap%2Fapp_bootstrap.go#L134-L136) in case the database initialization fails.

Well, in my case, the error message is following:
```
2025-12-09T12:34:11+02:00 INF cmd/root.go:118 > Starting Tinyauth version=development
2025-12-09T12:34:11+02:00 FTL cmd/root.go:128 > Failed to setup app error="failed to initialize database service: unable to open database file: out of memory (14)"
```

steveiliop56:
Yes, this is correct. That's the expected output of the driver when the database is not reachable. It will say out of memory which is a bit cryptic but at the same time it immediately points to the issue.

PR #509, opened 2025-12-10
LDAP: Add mTLS / client certificate authentication support

plaes:
It's now possible to connect to LDAP servers using client certificates. It was tested against Google's LDAP service (though only available for Google Workspace users):

```shell
export LDAP_ADDRESS="ldaps://ldap.google.com:636"
export LDAP_BASE_DN="dc=example,dc=com"
export LDAP_AUTH_CERT=Google_ldap_service.crt
export LDAP_AUTH_KEY=Google_ldap_service.key
# export LDAP_SEARCH_FILTER="(uid=%s)" ## <-- default works well for username-based users
```
When any of the `ldap-auth-*` variables is used, it will attempt to load certificate-key pair and if something fails, it exits immedately ignoring `ldap-bind-dn` and `ldap-bind-password` values.

Also added note about missing STARTTLS support.

## Summary by CodeRabbit

* **New Features**
  * LDAP authentication now supports mutual TLS (mTLS) using a client certificate and key.
  * LDAP configuration adds fields to specify client certificate and key paths.
  * LDAP service account binding updated to prefer an mTLS-capable binding flow when a client certificate is provided.

### DIFF
--- a/internal/service/ldap_service.go
+++ b/internal/service/ldap_service.go
@@ -87,7 +87,7 @@ func (ldap *LdapService) connect() (*ldapgo.Conn, error) {
 	var err error
 
 	// TODO: There's also STARTTLS (or SASL)-based mTLS authentication
-	// scenarion, where we first connect to plain text port (389) and
+	// scenario, where we first connect to plain text port (389) and
 	// continue with a STARTTLS negotiation:
 	// 1. conn = ldap.DialURL("ldap://ldap.example.com:389")
 	// 2. conn.StartTLS(tlsConfig)


internal/service/ldap_service.go
```diff
@@ -19,21 +19,34 @@ type LdapServiceConfig struct {
...
 func (ldap *LdapService) Init() error {
+	// Check whether authentication with client certificate is possible
+	if ldap.config.AuthCert != "" && ldap.config.AuthKey != "" {
+		cert, err := tls.LoadX509KeyPair(ldap.config.AuthCert, ldap.config.AuthKey)
+		if err != nil {
+			log.Fatal().Err(err).Msg("LDAP mTLS authentication")
+		} else {
+			ldap.cert = &cert
+			log.Info().Msg("Using LDAP mTLS authentication")
+		}
```
steveiliop56, line 45: I agree with @coderabbitai here. Instead of failing the app from within the initiliazation function, it is better to return the error. So, something like:

```go
if err != nil {
  return err
}
ldap.cert = &cert
log.Info().Msg("Using LDAP mTLS authentication")
```

internal/service/ldap_service.go
```diff
@@ -107,6 +135,20 @@ func (ldap *LdapService) Search(username string) (string, error) {
...
+	if rebind {
+		ldap.mutex.Lock()
+		defer ldap.mutex.Unlock()
+	}
+
+	if ldap.cert != nil {
+		return ldap.conn.ExternalBind()
+	} else {
+		return ldap.conn.Bind(ldap.config.BindDN, ldap.config.BindPassword)
+	}
```
steveiliop56, line 157: The else statement is not needed. This can be rewritten as:

```go
if ldap.cert != nil {
  return ldap.conn.ExternalBind()
}
return ldap.conn.Bind(ldap.config.BindDN, ldap.config.BindPassword)
```

plaes:
PR has been rebase against main and tested it against Google LDAP service.

Though there seem to be some regressions due to 03ed18343ea306aea4d7e5d65e50e6a993fe3351:
1. Configuration doesn't seem to be picked up from environment variables anymore (few examples: `TINYAUTH_APPURL` or `TINYAUTH_LDAP_ADDRESS` nor others seem to be working for me)
2. Default searchfilter for LDAP is empty now. Previously it defaulted to `(uid=%s)`.

steveiliop56:
@plaes no worries, I will look into them. Thank you for the pull request.

internal/service/ldap_service.go
```diff
@@ -60,31 +72,46 @@ func (ldap *LdapService) connect() (*ldapgo.Conn, error) {
...
+	// TODO: There's also STARTTLS (or SASL)-based mTLS authentication
+	// scenarion, where we first connect to plain text port (389) and
+	// continue with a STARTTLS negotiation:
+	// 1. conn = ldap.DialURL("ldap://ldap.example.com:389")
+	// 2. conn.StartTLS(tlsConfig)
+	// 3. conn.externalBind()
+	if ldap.cert != nil {
+		conn, err = ldapgo.DialURL(ldap.config.Address, ldapgo.DialWithTLSConfig(&tls.Config{
+			MinVersion:   tls.VersionTLS12,
+			Certificates: []tls.Certificate{*ldap.cert},
```
steveiliop56, line 98: Should we use the `InsecureSkipVerify` here too? Or will certificates handle this?

steveiliop56:
@plaes fixed the second issue you mentioned in b159f4472902eba1dec9461e713dcbb73de5803c, forgot to set the default. As for the first issue are you sure you configured everything correctly? I cannot reproduce. I also left you a small comment.

internal/service/ldap_service.go
plaes: I would rather keep it the way it is currently as `InsecureSkipVerify` is a huge security footgun with mTLS.
In case of "untrusted" certificates, one could populate the `tls.Config.RootCAs` with required CA certificate(s):
```go
    // Load CA cert(s)
    caCert, err := ioutil.ReadFile(*caFile)
    if err != nil {
        log.Fatal(err)
    }
    caCertPool := x509.NewCertPool()
    caCertPool.AppendCertsFromPEM(caCert)

    tlsConfig := &tls.Config{
        Certificates: []tls.Certificate{cert},
        RootCAs:      caCertPool,
    }
```

plaes: I added extra TODO note about that for now.

steveiliop56 (approved):

internal/service/ldap_service.go
steveiliop56: Alright, let's keep it this way for now.

PR #533, opened 2025-12-21
feat: unified config

steveiliop56:
## Summary by CodeRabbit

* **New Features**
  * YAML configuration example and namespaced TINYAUTH_ environment variables
  * New CLI entrypoint with subcommands for create/generate/verify/healthcheck/version
  * Multiple config loaders: file, flags, and environment

* **Improvements**
  * Reorganized hierarchical config (Server, Auth, OAuth, UI, LDAP)
  * Docker image/runtime env and healthcheck updates
  * Build workflows now target the CLI package

* **Chores**
  * Cleaned/updated ignores and dependency adjustments

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -16,8 +16,8 @@ COPY ./air.toml ./
 
 EXPOSE 3000
 
-ENV DATABASEPATH=/data/tinyauth.db
+ENV TINYAUTH_DATABASEPATH=/data/tinyauth.db
 
-ENV RESOURCESDIR=/data/resources
+ENV TINYAUTH_RESOURCESDIR=/data/resources
 
 ENTRYPOINT ["air", "-c", "air.toml"]
--- a/Dockerfile.distroless
+++ b/Dockerfile.distroless
@@ -55,9 +55,9 @@ EXPOSE 3000
 
 VOLUME ["/data"]
 
-ENV DATABASEPATH=/data/tinyauth.db
+ENV TINYAUTH_DATABASEPATH=/data/tinyauth.db
 
-ENV RESOURCESDIR=/data/resources
+ENV TINYAUTH_RESOURCESDIR=/data/resources
 
 ENV GIN_MODE=release
 
--- a/cmd/tinyauth/healthcheck.go
+++ b/cmd/tinyauth/healthcheck.go
@@ -29,14 +29,14 @@ func healthcheckCmd() *cli.Command {
 		Run: func(args []string) error {
 			log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr, TimeFormat: time.RFC3339}).With().Caller().Logger().Level(zerolog.InfoLevel)
 
-			appUrl := os.Getenv("APPURL")
+			appUrl := os.Getenv("TINYAUTH_APPURL")
 
 			if len(args) > 0 {
 				appUrl = args[0]
 			}
 
 			if appUrl == "" {
-				return errors.New("APPURL is not set and no argument was provided")
+				return errors.New("TINYAUTH_APPURL is not set and no argument was provided")
 			}
 
 			log.Info().Str("app_url", appUrl).Msg("Performing health check")


steveiliop56:
@coderabbitai review

cmd/healthcheck.go
```diff
@@ -1,112 +1,83 @@
...
-	if c.root != nil {
-		c.root.AddCommand(c.cmd)
-	}
-}
+			log.Info().Str("app_url", appUrl).Msg("Performing health check")
 
-func (c *healthcheckCmd) GetCmd() *cobra.Command {
-	return c.cmd
-}
+			client := http.Client{}
```
Rycochet: I'd agree with coderabbit here - add a timeout, if this doesn't return reasonably quickly then there's a problem that should probably cause a restart if it persists - I'd almost suggest something smaller like 5 seconds (given that there need to be 3 successive fails to mark a container as unhealthy)

```suggestion
			client := http.Client{
				Timeout: 30 * time.Second,
			}
```

internal/config/config.go
```diff
@@ -15,36 +15,67 @@ var RedirectCookieName = "tinyauth-redirect"
...
-	LdapInsecure          bool   `mapstructure:"ldap-insecure"`
-	LdapSearchFilter      string `mapstructure:"ldap-search-filter"`
-	ResourcesDir          string `mapstructure:"resources-dir"`
-	DatabasePath          string `mapstructure:"database-path"`
-	TrustedProxies        string `mapstructure:"trusted-proxies"`
-	DisableAnalytics      bool   `mapstructure:"disable-analytics"`
-	DisableResources      bool   `mapstructure:"disable-resources"`
-	DisableUIWarnings     bool   `mapstructure:"disable-ui-warnings"`
-	SocketPath            string `mapstructure:"socket-path"`
+	AppURL            string             `description:"The base URL where the app is hosted." yaml:"appUrl"`
```
Rycochet, line 18: Generic question on the yaml names - this is camelCase, but the standard in say docker compose files is snake_case - does traefik config translate for both, or is it case insensitive, or is this a deliberate decision?

.env.example
```diff
@@ -1,22 +1,86 @@
...
-RESOURCES_DIR=/data/resources
-DATABASE_PATH=/data/tinyauth.db
-DISABLE_ANALYTICS=false
-DISABLE_RESOURCES=false
-TRUSTED_PROXIES=
\ No newline at end of file
+# Base Configuration
+
+# The base URL where Tinyauth is accessible
+TINYAUTH_APPURL=https://auth.example.com
```
Rycochet: Semi-echoing coderabbit here - but personal preference due to inconsistencies around the OS world - quote all values for consistency (even if ones with spaces are the only ones that really have problems anywhere)

.env.example
```diff
@@ -1,22 +1,86 @@
...
+# Path to external users file (optional)
+TINYAUTH_USERSFILE=
+# Enable secure cookies (requires HTTPS)
+TINYAUTH_SECURECOOKIE=true
+# Session expiry in seconds (7200 = 2 hours)
+TINYAUTH_SESSIONEXPIRY=7200
+# Login timeout in seconds (300 = 5 minutes)
+TINYAUTH_LOGINTIMEOUT=300
+# Maximum login retries before lockout
+TINYAUTH_LOGINMAXRETRIES=5
```
Rycochet: General comment (future PR - but I don't want to forget to say) - if there's a lockout, what happens? Maybe needs an option for how long the IP is locked out for? Is it by IP, mask, session (though that's not secure), username?

.env.example
```diff
@@ -1,22 +1,86 @@
...
+# OAuth Configuration
+
+# Regex pattern for allowed email addresses (e.g., /@example\.com$/)
+TINYAUTH_OAUTH_WHITELIST=
+# Provider ID to auto-redirect to (skips login page)
+TINYAUTH_OAUTH_AUTOREDIRECT=
+# OAuth Provider Configuration (replace MYPROVIDER with your provider name)
+TINYAUTH_OAUTH_PROVIDERS_MYPROVIDER_CLIENTID=your_client_id_here
+TINYAUTH_OAUTH_PROVIDERS_MYPROVIDER_CLIENTSECRET=your_client_secret_here
+TINYAUTH_OAUTH_PROVIDERS_MYPROVIDER_AUTHURL=https://provider.example.com/oauth/authorize
```
Rycochet: Could probably add a comment to these three stating they're required (and noting google / github don't need url) - also blank line before 53 to separate them from the non-provider options?

config.example.yaml
```diff
@@ -0,0 +1,88 @@
...
+  # Regex pattern for allowed email addresses (e.g., /@example\.com$/)
+  whitelist: ""
+  # Provider ID to auto-redirect to (skips login page)
+  autoRedirect: ""
+  # OAuth Provider Configuration (replace myprovider with your provider name)
+  providers:
+    myprovider:
+      clientId: "your_client_id_here"
+      clientSecret: "your_client_secret_here"
+      authUrl: "https://provider.example.com/oauth/authorize"
```
Rycochet, line 57: Echoing comment in ENV about adding comments to these three required lines

internal/config/config.go
steveiliop56: We need camel-case for things like the CLI flags and environment variables.

.env.example
steveiliop56: If you fail to login 5 times, the login endpoint simply denies authentication. The way it works is by storing the username and essentially locking the account for the timeout.

config.example.yaml
steveiliop56: This will be mentioned in the documentation, the example file..is just an example.

.env.example
steveiliop56: See the other comment.

> This will be mentioned in the documentation, the example file..is just an example.

PR #550, opened 2025-12-30
feat: make tinyauth an oidc provider

steveiliop56:


steveiliop56:
@shreknel after spending some time reviewing this pull request, I will have to close it. This is because it involves a very sensitive aspect of a soon to become feature of Tinyauth.

I found multiple issues that do not follow the project's quality standards. It seems that you have created this pull request largely using the assistance of an LLM and thus I cannot trust the output of an algorithm enough to merge into the main branch especially with something this complex.
 
However, this doesn't mean that the feature will not make it into Tinyauth. This pull request gave me some ideas of how to implement OIDC which I will be doing very soon. I believe it's better to start from scratch rather than try to review hundreds of lines of code and fix countless issues.

PR #627, opened 2026-02-04
feat(access-control): Add support for Kubernetes Label

contre95:
## Release Notes

* **New Features**
  * Added support for reading application configuration from Kubernetes Ingress annotations in addition to Docker container labels.

* **Documentation**
  * Updated README to document the new Kubernetes Ingress annotation configuration option.
  * Added support for reading application configuration from Kubernetes Ingress annotations.
  * Added `LabelProvider` configuration option to choose between Docker labels and Kubernetes for ACL configuration.

## Summary by CodeRabbit

* **New Features**
  * Added support for Kubernetes as a label provider for access control configurations, enabling ACL settings to be sourced from Kubernetes Ingresses as an alternative to Docker-based providers.


contre95:
Hi @steveiliop56, before I continue with this feature. I wanted to check if this is something you wanted in Tinyauth. I also wanted to ask if I should part from `main` 🏷️ `nightly` or from `main` 🏷️ `v4.1.0` (or latest version). Thanks and great authenticator.

contre95:
@coderabbitai review

steveiliop56:
@contre95 thank you for the pull request. From a quick look it seems good but I will unfortunately have to hold it until the next version of Tinyauth (`v5.1.0`). Version 5 already has a ton of things that need testing and when I finish it up we can proceed with the new features. Ideally you should have the `main` branch as the base to get the most recent changes.

contre95:
@steveiliop56 No problem, I bet it can wait. I've been testing `:nightly` and it seems that many things changed, make sense. There are some things I'd like to improve still, so there's no rush.

internal/bootstrap/service_bootstrap.go
scottmckendry: `dockerService` can be `nil` here when label provider is k8s. Based on the rest of the PR, this param is no longer used, so should probably be removed from `AuthService`/`NewAuthService`

internal/service/kubernetes_service.go
```diff
@@ -0,0 +1,490 @@
...
+func (k *KubernetesService) Init() error {
+	var config *rest.Config
+	var err error
+
+	// Try in-cluster config first
+	config, err = rest.InClusterConfig()
+	if err != nil {
+		tlog.App.Error().Err(err).Msg("Failed to get in-cluster Kubernetes config")
+		k.started = false
+		return nil
```
scottmckendry, line 285: All of the error paths in `Init()` return `nil` instead of a descriptive error. 

service_bootstrap.go has logic to handle errors returned by this method that will never get used.

internal/service/kubernetes_service.go
```diff
@@ -0,0 +1,490 @@
...
+	defer resyncTicker.Stop()
+
+	// Initial resync
+	if err := k.resyncGVR(gvr); err != nil {
+		tlog.App.Error().Err(err).Str("api", gvr.GroupVersion().String()).Msg("Initial resync failed, retrying in 30 seconds")
+		time.Sleep(30 * time.Second)
+	}
+
+	for {
+		select {
```
scottmckendry, line 187: If the first two cases here fall through and we hit `default`, but `watcher` does not return an error, it could create a continuous loop.

I think we should add a small sleep after the loop exits, regardless of any of the conditions here (i.e. not just if watcher returns an error)

internal/service/kubernetes_service.go
```diff
@@ -0,0 +1,490 @@
...
+			go k.watchIngressV1()
+		} else {
+			tlog.App.Warn().Msg("Insufficient permissions for networking.k8s.io/v1 Ingress")
+			v1Available = false
+		}
+	} else {
+		tlog.App.Debug().Msg("networking.k8s.io/v1 Ingress API not available")
+	}
+
+	if v1beta1Available {
```
scottmckendry, line 367: pretty sure this was removed in k8s v1.22 - I don't think we should be worried about supporting any versions of kubernetes older than this - so we can probably remove this var and it's logic entirely. 

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

internal/service/kubernetes_service.go
```diff
@@ -0,0 +1,490 @@
...
+			tlog.App.Debug().Err(err).Msg("Failed to get labels from v1beta1 Ingress")
+		}
+		if app.Config.Domain != "" {
+			return app, nil
+		}
+	}
+	return config.App{}, nil
+}
+
+func (k *KubernetesService) getLabelsFromGVR(gvr schema.GroupVersionResource, appDomain string) (config.App, error) {
```
scottmckendry, line 438: if the watch goroutine and the 5-minute resync are already responsible for keeping the cache populated, why does `GetLabels` also make a live API call to Kubernetes on every cache miss?

This may be redundant... correct me if I am wrong 🙂

internal/service/kubernetes_service.go
```diff
@@ -0,0 +1,490 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+)
+
+var _ = unstructured.Unstructured{}
```
scottmckendry, line 22: I don't think we need this line. compiles locally for me without it 👍🏻

internal/service/kubernetes_service.go
scottmckendry, line 1: We probably need at least a couple tests for this service.

scottmckendry (changes requested):
A few minor things I noticed. I think the main thing missing is test coverage, IMO. Otherwise looks pretty good 🙂

contre95:
@scottmckendry Thanks for the comments. All valid. I'll try to get work on them before 5.1.0 which I believe is the version that will include my change.

internal/service/kubernetes_service.go
contre95: Hey, good catch.
 
I was very hesitant at first to add a watcher vs just periodic syncs, cause I knew I would mess something up. I was wondering if a Watcher and periodic re-sync are both necessary or if it might make the code more complicate it. 
Real time updates are neat tho, and those come from the watcher. At the moment, I'm running my custom version of Tinyauth in my homelab k8s and changing permission on annotation and seeing them implemented by default it really neat. Also useful for debugging this very same feature.

internal/service/kubernetes_service.go
contre95: You are right, I left this from when I had no watcher, thanks !

contre95:
Hey @scottmckendry thanks for the feedback, I think I have solved all of the comments.
@steveiliop56 do you think this will still make it to `5.1.0` ?

internal/service/kubernetes_service.go
```diff
@@ -0,0 +1,383 @@
...
+				if app.appName == appName {
+					return app.app, true
+				}
+			}
+		}
+	}
+	return config.App{}, false
+}
+
+func (k *KubernetesService) updateFromList(items []unstructured.Unstructured) {
```
scottmckendry: The logic in this function is almost identical to `case watch.Added, watch.Modified:` in `watchGVR`

Could be worth breaking out.

internal/service/kubernetes_service.go
```diff
@@ -0,0 +1,383 @@
...
+	if err != nil {
+		tlog.App.Debug().Err(err).Str("api", gvr.GroupVersion().String()).Msg("Failed to list ingresses during resync")
+		return err
+	}
+	k.updateFromList(list.Items)
+	tlog.App.Debug().Str("api", gvr.GroupVersion().String()).Int("count", len(list.Items)).Msg("Resynced ingress cache")
+	return nil
+}
+
+func (k *KubernetesService) watchGVR(gvr schema.GroupVersionResource) {
```
scottmckendry, line 206: This function feels a little lengthy. My previous comment RE: duplicated logic may help.

But I also think it's worth splitting into two - one function to handle the outer for loop, ticker etc. and another to handle each instance of the "watch". If that makes sense?

Right now it's a bit challenging to follow at a glance.

scottmckendry (dismissed):
Good to go. Two minor comments that I think are worth looking at but I'm happy to approve.

internal/service/kubernetes_service.go
contre95: You are right, I'll use `updateFromList` directly in watchGVR and loop over the different items on rsyncGVR

internal/service/kubernetes_service.go
contre95: Fair, I just added a `runWatcher` .

contre95:
> Good to go. Two minor comments that I think are worth looking at but I'm happy to approve.

Done !

internal/bootstrap/service_bootstrap.go
```diff
@@ -38,17 +39,30 @@ func (app *BootstrapApp) initServices(queries *repository.Queries) (Services, er
...
+		labelProvider = kubernetesService
+	default:
+		dockerService = service.NewDockerService()
+		err = dockerService.Init()
+		if err != nil {
+			return Services{}, err
+		}
+		services.dockerService = dockerService
+		labelProvider = dockerService
 	}
```
steveiliop56, line 69: Can't we automatically detect the ACLs provider perhaps with an `auto` value?

internal/service/kubernetes_service.go
```diff
@@ -0,0 +1,354 @@
...
+type ingressAppKey struct {
+	ingressKey
+	appName string
+}
+
+type ingressApp struct {
+	domain  string
+	appName string
+	app     config.App
+}
```
steveiliop56, line 36: Can we please capitalize the first letter of types? Just to avoid confusion with types and variables.

internal/service/kubernetes_service_test.go
steveiliop56, line 1: Can we please follow the testing scheme of the controllers? It makes testing easier since you don't have to initialize everything again.

internal/service/label_provider.go
steveiliop56, line 1: I believe we can throw this in the access controls service, no need for a separate file.

internal/service/kubernetes_service.go
```diff
@@ -0,0 +1,354 @@
...
+	v1Available := false
+	for _, apiGroup := range apiGroups {
+		if apiGroup.GroupVersion == "networking.k8s.io/v1" {
+			for _, resource := range apiGroup.APIResources {
+				if resource.Name == "ingresses" && resource.Kind == "Ingress" {
+					v1Available = true
+				}
+			}
+		}
+	}
```
steveiliop56: Is there a specific reason we look for v1 or is it just to prevent breaking changes?

steveiliop56 (changes requested):
@contre95 some nitpicks and we should be good to go.

> @steveiliop56 do you think this will still make it to 5.1.0 ?

Yes absolutely!

internal/service/kubernetes_service.go
contre95: I think this check made sense from when we were supporting `v1' and `v1beta1` for K8S 1.22.
We can remove the whole api version discovery process and leave it just to `v1` this will make code much simpler. I'll try do it toda afternoon and test it with and older cluster I have  to check behavior. I wouldn't want it fail silently on al older clusters.

internal/service/kubernetes_service.go
contre95: This sound a bit weird. This would export the variables that are only withn the kubernetes_service.go. I don't see any other part of Tinyauth that should use these and I believe that not exporting them would avoid confusion on the LSP of someone tempering the lables in other part of the system. Am I missing something ? 
But it's your call. Happy to make change.

internal/bootstrap/service_bootstrap.go
contre95: I like the approach, less  configuration for the end user and it aligns with Golangs

contre95: After a quick chat with AI, apparently the best way to do this is to check the `KUBERNETES_SERVICE_HOST`. I will do that and default to docker if not present.

internal/service/label_provider.go
contre95: Done 👍🏼

internal/service/kubernetes_service.go
steveiliop56: @contre95 forgot this one.

contre95: Sorry, I always forget to hit 'Sumbit' when I comment on other ppl comments. 
wdyt about https://github.com/tinyauthapp/tinyauth/pull/627#discussion_r3126391267 ?

steveiliop56 (approved):
lgtm

internal/service/kubernetes_service.go
steveiliop56: It's probably just a habit of mine. I use `Foo` as the struct name and then do `foo := Foo{}` but eh it's probably a me thing.

scottmckendry (approved):

steveiliop56:
Thanks @contre95!

nebula-it:
@contre95 Thank you for adding this feature. Just wanted to check and see if there is any documentation for this? I wasn't able to find anything [here ](https://tinyauth.app/docs/community/kubernetes/)

PR #629, opened 2026-02-05
fix(frontend): Add flag to check auto-redirect

contre95:
The [useEffect](https://github.com/contre95/tinyauth/blob/3bf8c62542c6204518c4e5543110df723abdacf5/frontend/src/pages/login-page.tsx?plain=1#L125) hook had a dependency on [oauthMutation](https://github.com/contre95/tinyauth/blob/3bf8c62542c6204518c4e5543110df723abdacf5/frontend/src/pages/login-page.tsx?plain=1#L136), which changes every time the mutation runs. This caused the effect to re-run indefinitely, creating hundreds of requests to  the OIDC provider, in this case`/api/oauth/url/pocketid`.

This is happening only in `tinyauth:nightly` 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of OAuth auto-redirect flow during login to prevent duplicate redirect attempts.
  * Enhanced error handling to ensure proper recovery from OAuth authentication failures.

### DIFF
--- a/frontend/src/pages/login-page.tsx
+++ b/frontend/src/pages/login-page.tsx
@@ -46,6 +46,7 @@ export const LoginPage = () => {
 
   const redirectTimer = useRef<number | null>(null);
   const redirectButtonTimer = useRef<number | null>(null);
+  const hasAutoRedirected = useRef(false);
 
   const searchParams = new URLSearchParams(search);
   const {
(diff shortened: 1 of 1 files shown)


steveiliop56:
Hey @contre95, thank you for the pull request. I will have to close this one because I am already working on a small refactor for all of the effect calls in #630. In any case thank you for the `useRef` idea, it does solve a lot of problems!

contre95:
@steveiliop56 No problem !  Just in case, I realized about this issue with the `nightly` image while trying to access a service behind Tinyauth while not being authenticated.

steveiliop56:
Yes it was one of the many issues in the alpha/nightly *lol*.

PR #640, opened 2026-02-11
feat: reload users file on SIGHUP signal

frankhommers:
## Summary

Add a SIGHUP signal handler that re-reads the users file (`auth.usersFile`) and updates the in-memory user list without requiring a full restart.

## Motivation

Currently, any change to the users file requires a full container/process restart. This makes it difficult for external tools (e.g. user management sidecars, scripts) to manage users dynamically.

With this change, after modifying the users file you can simply send:
```bash
kill -HUP $(pidof tinyauth)
# or from outside the container:
docker exec tinyauth kill -HUP 1
```

## Changes

- **`internal/service/auth_service.go`**:
  - Add `usersMutex sync.RWMutex` for thread-safe access to the users slice
  - Add `ReloadUsers()` method that swaps the users list under a write lock
  - Protect `GetLocalUser()` and `LocalAuthConfigured()` with read locks

- **`internal/bootstrap/app_bootstrap.go`**:
  - Add `handleSIGHUP()` goroutine that listens for SIGHUP and calls `ReloadUsers()`
  - Re-reads both inline users (`auth.users`) and users file (`auth.usersFile`)

## Notes

- Zero impact on existing behavior — without SIGHUP, nothing changes
- Minimal diff: 38 lines added, 2 files changed
- Thread-safe: RWMutex ensures no race conditions during reload
- Follows existing patterns (similar to `dbCleanup` goroutine)

## Summary by CodeRabbit

* **New Features**
  * Added runtime user reload via SIGHUP so users can be refreshed without restarting the application.

* **Improvements**
  * Safer in-memory user handling with concurrency protection to prevent race conditions.
  * Authentication provider configuration now refreshes and logs state so UI reflects updated provider/identity sources after reloads.

### DIFF
--- a/internal/bootstrap/app_bootstrap.go
+++ b/internal/bootstrap/app_bootstrap.go
@@ -145,43 +145,9 @@ func (app *BootstrapApp) Setup() error {
 	app.services = services
 
 	// Configured providers
-	configuredProviders := make([]controller.Provider, 0)
-
-	for id, provider := range app.context.oauthProviders {
-		configuredProviders = append(configuredProviders, controller.Provider{
-			Name:  provider.Name,
-			ID:    id,
-			OAuth: true,
-		})
+	if err := app.refreshConfiguredProviders(); err != nil {
+		return err
 	}
-
-	sort.Slice(configuredProviders, func(i, j int) bool {
-		return configuredProviders[i].Name < configuredProviders[j].Name
-	})
-
-	if services.authService.LocalAuthConfigured() {
-		configuredProviders = append(configuredProviders, controller.Provider{
-			Name:  "Local",
-			ID:    "local",
-			OAuth: false,
-		})
-	}
-
-	if services.authService.LdapAuthConfigured() {
-		configuredProviders = append(configuredProviders, controller.Provider{
-			Name:  "LDAP",
-			ID:    "ldap",
-			OAuth: false,
-		})
-	}
-
-	tlog.App.Debug().Interface("providers", configuredProviders).Msg("Authentication providers")
-
-	if len(configuredProviders) == 0 {
-		return fmt.Errorf("no authentication providers configured")
-	}
-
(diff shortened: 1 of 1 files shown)


steveiliop56:
@frankhommers thank you for the pull request. I am a bit skeptical about it though. I am thinking what's the advantage of handling SIGHUP if the user has to execute a command to trigger it? I mean if the user has to run a command they might as well restart Tinyauth which is super fast anyway. Maybe we should instead add a file watcher? Maybe a routine that watches the user file (and maybe even the secret files for OAuth providers/LDAP etc) and just reloads the configured providers and service secrets?

frankhommers:
The problem is that the health check takes too long, before it's up again, and traefik in turn will give a 404. 
A SIGHUP wouldn't interfere with the health status.

Filewatchers would be even better yes! That would be awesome (for all configs)!  But I think we should put that into v5 instead?

But it also gives me an idea to make the health check quicker ;-)

How should we proceed?

steveiliop56:
I am going to release v5 today so, I am not adding any new features that need additional testing.

I believe we can't make the health-check any faster since it's just a web request so it relies on the frequency the user defines.

As for the file watcher I believe it's the right way to implement such feature. Would you like to work on something like this? If so, should it be on another pull request or in this one?

frankhommers:
I was thinking of making a health check script with a higher interval and doing `no-ops` most of the time.
Anyway, let's close this PR. 

Filewatchers is a different thing, and it makes no sense to continue building on this PR, because it needs to be reverted anyway ;-)

steveiliop56:
In any case thank you for your time working on this @frankhommers. If you like, I could ping you when I start working on the file watcher. Will take a bit though since we need to firstly move the config file option out of experimental.

frankhommers:
Maybe it's better to wait on that. Ideally we would watch that config file too ;-)

PR #696, opened 2026-03-07
feat: add redis session driver

henrywhitaker3:
Adds a new session driver option to add redis as a backend for sessions. To use it, set the following vars:

- `TINYAUTH_DATABASE_REDIS_URL=redis.redis.svc:6379`
- `TINYAUTH_SESSION_DRIVER=redis`

## Summary by CodeRabbit

* **New Features**
  * Configurable Redis session support (URL/password/DB) with default driver remaining "database".
  * Redis-backed session store: create, retrieve, update, and delete sessions in Redis.
  * Authentication now uses the configured session backend so existing auth flows work with Redis sessions.

### DIFF
--- a/internal/bootstrap/app_bootstrap.go
+++ b/internal/bootstrap/app_bootstrap.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/redis/rueidis"
 	"github.com/steveiliop56/tinyauth/internal/config"
 	"github.com/steveiliop56/tinyauth/internal/controller"
 	"github.com/steveiliop56/tinyauth/internal/repository"
@@ -125,16 +126,28 @@ func (app *BootstrapApp) Setup() error {
 
 	// Database
 	db, err := app.SetupDatabase(app.config.Database.Path)
-
 	if err != nil {
 		return fmt.Errorf("failed to setup database: %w", err)
+	}
+
+	var redisClient rueidis.Client
+	if app.config.Session.Driver == "redis" {
+		client, err := app.SetupRedis(
+			app.config.Database.Redis.URL,
+			app.config.Database.Redis.Password,
+			app.config.Database.Redis.DB,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to setup redis client: %w", err)
+		}
+		redisClient = client
 	}
 
 	// Queries
 	queries := repository.New(db)
 
 	// Services
-	services, err := app.initServices(queries)
+	services, err := app.initServices(queries, redisClient)
 
 	if err != nil {
 		return fmt.Errorf("failed to initialize services: %w", err)
--- a/internal/bootstrap/service_bootstrap.go
+++ b/internal/bootstrap/service_bootstrap.go
@@ -1,6 +1,9 @@
 package bootstrap
 
 import (
+	"fmt"
+
+	"github.com/redis/rueidis"
 	"github.com/steveiliop56/tinyauth/internal/repository"
 	"github.com/steveiliop56/tinyauth/internal/service"
 	"github.com/steveiliop56/tinyauth/internal/utils/tlog"
@@ -15,7 +18,7 @@ type Services struct {
 	oidcService          *service.OIDCService
 }
 
-func (app *BootstrapApp) initServices(queries *repository.Queries) (Services, error) {
+func (app *BootstrapApp) initServices(queries *repository.Queries, redis rueidis.Client) (Services, error) {
 	services := Services{}
 
 	ldapService := service.NewLdapService(service.LdapServiceConfig{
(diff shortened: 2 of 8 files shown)


internal/bootstrap/app_bootstrap.go
```diff
@@ -125,16 +126,28 @@ func (app *BootstrapApp) Setup() error {
...
+		client, err := app.SetupRedis(
+			app.config.Database.Redis.URL,
+			app.config.Database.Redis.Password,
+			app.config.Database.Redis.DB,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to setup redis client: %w", err)
+		}
+		redisClient = client
+	}
```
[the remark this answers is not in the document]
henrywhitaker3: Added a TTL for redis records, there's no db.Close() either for the existing storage, so haven't added that for redis

internal/bootstrap/service_bootstrap.go
```diff
@@ -58,6 +61,16 @@ func (app *BootstrapApp) initServices(queries *repository.Queries) (Services, er
...
 
+	var sessionRepo service.SessionRepository
+	switch app.config.Session.Driver {
+	case "database":
+		sessionRepo = queries
+	case "redis":
+		sessionRepo = repository.NewRedisSessionRepository(redis)
+	default:
+		return Services{}, fmt.Errorf("unknown session driver %s", app.config.Session.Driver)
+	}
```
steveiliop56, line 72: In the future we may need more session drivers and this approach would get complicated fast. I recommend moving all query logic in a query service that automatically returns the correct query based on the user choice.

internal/bootstrap/service_bootstrap.go
```diff
@@ -70,7 +83,7 @@ func (app *BootstrapApp) initServices(queries *repository.Queries) (Services, er
 		SessionCookieName:  app.context.sessionCookieName,
 		IP:                 app.config.Auth.IP,
 		LDAPGroupsCacheTTL: app.config.Ldap.GroupCacheTTL,
-	}, dockerService, services.ldapService, queries)
+	}, dockerService, services.ldapService, sessionRepo)
```
steveiliop56, line 86: If we add a Redis driver, we need to utilize it for everything. This, in combination with https://github.com/steveiliop56/tinyauth/pull/696/changes#r3046330528 should provide all services/controllers that need persistent storage the correct query.

internal/config/config.go
```diff
@@ -79,6 +82,7 @@ var RedirectCookieName = "tinyauth-redirect"
 type Config struct {
 	AppURL       string             `description:"The base URL where the app is hosted." yaml:"appUrl"`
 	Database     DatabaseConfig     `description:"Database configuration." yaml:"database"`
+	Session      SessionConfig      `description:"Session configuration" yaml:"session"`
```
steveiliop56, line 85: We don't really need a config option. We just need to check if Redis is configured. If it is, we use it, if not we fallback to SQLite.

internal/repository/redis_session_queries.go
steveiliop56, line 1: We need a way to generalize these functions to handle any data. This way they can be utilized by a query service as mentioned in https://github.com/steveiliop56/tinyauth/pull/696/changes#r3046330528 and store all sorts of data. We have two options as to how to approach this, either we wait for [generics in methods](https://github.com/golang/go/issues/77273) or we move the set/get/edit in smaller handler functions that just take bytes as props and we handle the serialization/deserialization in some helper function in the queries service.

internal/service/auth_service.go
```diff
@@ -52,10 +61,10 @@ type AuthService struct {
 	loginMutex      sync.RWMutex
 	ldapGroupsMutex sync.RWMutex
 	ldap            *LdapService
-	queries         *repository.Queries
+	queries         SessionRepository
```
steveiliop56, line 64: This should be handled by the queries service which should provide all of the queries from the repository.

steveiliop56 (commented):
@henrywhitaker3 this is a cool addition but it requires a huge refactor in the way queries are handled.

We basically need to move the query handling from the SQLite repository to a service that will decide what backend to use for storing the data while ideally implementing the already existing queries interface to avoid refactors in the services and controllers. Essentially we need a drop in replacement for the queries that will handle the backends. While I do have a way in mind to implement such system (comments point to a possible implementation) it will require a fair amount of effort. I left the comments as mostly guidance for when we pick this up again.

If you like, you can try implementing such system and I will provide as much feedback as I can. Otherwise, I can takeover and move this to the backlog.

@Rycochet if you can, please share your thoughts.

Rycochet:
My thoughts were largely discussed on Discord - briefly "love the idea, think it's more complicated than it appears" 😅

Having the DB access abstracted, so we've got the built-in sqlite, then an option for redis (ie, this one), or postgres (or others if people are interested in adding) would make life better for setups where the sqlite deficiencies are more apparent (ie, a *lot* of users / connections).

Having this purely for caching would improve performance, but I think the maintainability of the code goes down with more code paths (not to say adding a cache service for it is a bad thing) - the actual stored data isn't that complicated so a key<>value store wouldn't be a problem for full use - then the options can be unified a bit more easily, and power users get more choice (in the future) ;-)

Rapter001:
@steveiliop56 @Rycochet @henrywhitaker3 there is also no name spaces for redis so if users decide to use 1 redis instance for all there apps it will conflict i have a list of thoughts

No startup connectivity check doesn’t verify Redis is up at startup.
No tls support.

PR #710, opened 2026-03-12
feat: add option to run tinyauth on a top-level domain

jacekkow:
It allows to set-up Tinyauth on top-level domain or to have it as a standalone OIDC server, without transparent authentication for apps in subdomains.

Additionally fix issue in release workflow where metadata (`VERSION`, `COMMIT_HASH`, `BUILD_TIMESTAMP`) is fetched from `nightly` tag rather than from the relevant `v...` tag.

## Summary by CodeRabbit

* **New Features**
  * Added an authentication setting to toggle subdomain vs standalone cookie scoping (defaults to enabled).
  * Session and OAuth cookie creation, refresh, and deletion now respect the configured mode so cookie scope behaves correctly for subdomains or single-host deployments.


steveiliop56:
#724 related

@jacekkow I believe the correct approach would be to automatically detect if we are running on a non-subdomain app URL. On start-up check what `strings.Split` returns. If it's 2 then we are running on the root, so we show a warning that cookies will be set for `.<root>` and continue as normal. If it's more than 2 we are running on a subdomain so we do the regular checks for the PSL and stuff.

jacekkow:
@steveiliop56 - that is not my use case. I want to use this app as a standalone component (auth server). I do not want it to set cookies for any other domain (esp. wildcard ones). Secured applications use OIDC and are at completely different URLs.

steveiliop56:
@jacekkow that's fine, it will not set cookies if you don't use the authentication feature. The idea is to just automatically handle root domains instead of needing another config option. Your pull request basically does this. Are you interested in implementing said feature or should I do it?

jacekkow:
> it will not set cookies if you don't use the authentication feature

OIDC login does set cookies.

>The idea is to just automatically handle root domains instead of needing another config option.

My instance of Tinyauth is **not** hosted on root domain!

So unless you've missed something my use case is different and requires a different solution than an algorithm from your first comment (https://github.com/steveiliop56/tinyauth/pull/710#issuecomment-4091456858).

steveiliop56:
> OIDC login does set cookies.

My brain was not thinking when I wrote this.

..for the rest..

Yes you are correct *again*, I thought your instance was on the root. Anyway in that case I guess we could have such option. But maybe under a different name? `SINGLEDOMAIN` is a bit confusing imo. Maybe something like `TINYAUTH_AUTH_COOKIEOIDCONLY`?

jacekkow:
@steveiliop56 Sure - naming can be changed.

What I also considered is having an option to just set arbitrary "cookie domain" (`TINYAUTH_AUTH_COOKIEDOMAIN` maybe) - and if it is not set default to the algorithm without any modifications. If wrong value is set, it's on the user :)

What would you prefer? Renaming "my option" to `TINYAUTH_AUTH_COOKIEOIDCONLY` (or maybe `TINYAUTH_AUTH_STANDALONE`?) or an option to set cookie domain?

steveiliop56:
I think the standalone idea is nice, let's change it to that. As for a cookie domain option, I think it's better to avoid adding it because it can *and probably will* cause confusion.

Also maybe we should add a small info log saying that authentication will not work for proxied apps with this option enabled (again to avoid future issues).

*Small note: This pull request is "queued" for v5.1.0 which will come later. I am focusing on fixing some issues in a patch release and then we can start adding new features.*

jacekkow:
@steveiliop56 Sure, no problem, I can work with a fork for now. BTW - would you like to have commit https://github.com/steveiliop56/tinyauth/pull/710/commits/646505756cb24f4e602b582d7c6a8443c7b924df submitted as a separate PR for the fix release?

steveiliop56:
Was planning to fix this with #715, I noticed it there too. But yeah feel free to create a pull request - please check if there is any nightly leftover anywhere else -.

jacekkow:
Haven't found any more of those. PR: https://github.com/steveiliop56/tinyauth/pull/725

Rycochet:
I don't like the naming of this option - it doesn't explain exactly what it does and has connotations that don't match, instead how about using the same pattern as analytics with a default-true value that explains it directly and call it `Subdomains` instead?

`TINYAUTH_AUTH_SUBDOMAINS` - and make it clear in the description (or docs) that this is "deeper subdomains" so it will work from whatever level you specify?

(The option itself is awesome though!)

jacekkow:
@steveiliop56 - should I do one more rename to `TINYAUTH_AUTH_SUBDOMAINS`?

steveiliop56:
@jacekkow I believe @Rycochet is correct standalone implies it runs without any additional software. How about `TINYAUTH_AUTH_SUBDOMAINSENABLED` which defaults to true? I think it makes the most sense?

*naming things is so unbelievably hard*

jacekkow:
I have changed the naming.

Coderabbit is actually correct that `cookieDomain` is always prefixed with `.` when setting the cookies. It is not an issue for me, but would you also like to have separate `appDomain` (used with `utils.CompileUserEmail`) and actual `cookieDomain` (set to `.{domain}` if subdomains are enables and otherwise just `{domain}`)?

steveiliop56:
@jacekkow if you can also fix that, it would be amazing. You should be able to do it by either passing the cookie domain already formatted so as the controllers don't have to append anything or by adding a separate option in the controller configuration. I would honestly prefer the first one.

jacekkow:
@steveiliop56 - thanks for taking that on - I'd have time only during next weekend.

BTW. Should maybe `CookieDomain` field in `AuthServiceConfig` be renamed to `MainDomain`, too?

jacekkow:
@steveiliop56  - thanks for taking that on.

BTW. Should `CookieDomain` in `AuthServiceConfig` be changed to `BaseDomain`/`MainDomain`, too? It has less to do with cookies now.

steveiliop56:
@jacekkow don't worry I will take it over from here, meaning we merge because nothing is left.

> BTW. Should maybe CookieDomain field in AuthServiceConfig be renamed to MainDomain, too?

Most probably yes. But let's leave it as-is for now since I would like to rework the cookie and context logic later.

steveiliop56 (dismissed):

internal/utils/app_utils.go
scottmckendry, line 16: Am I tripping or do we now have three implementations of `getCookieDomain`? This will cause a bug at some point in the future, I'm sure of it.

scottmckendry: Also, this one has test coverage, the other two do not.

internal/utils/app_utils.go
```diff
@@ -49,6 +49,15 @@ func GetCookieDomain(u string) (string, error) {
 	return domain, nil
 }
 
+func GetStandaloneCookieDomain(u string) (string, error) {
```
scottmckendry, line 50: This probably needs tests to cover the scenarios it may be used in.

scottmckendry (dismissed):

internal/utils/app_utils.go
steveiliop56: I mean do we really need testing? It's just URL parse and the stdlib is tested so what should we test? If it parses `example.com` and fails `example`? Because that's already covered.

internal/utils/app_utils.go
steveiliop56: The cookie domain/app URL stuff is in a bad state not gonna lie. It's confusing throughout the app but I believe we should rework them in another refactor rather than this pull request.

internal/utils/app_utils.go
scottmckendry: my take is that if it's a straightforward test to write, it's worth having. Especially if it's edge case config. Better to catch a regression in a test that have a user report their tinyauth is broken. Your call, though.

Rycochet (approved):

PR #727, opened 2026-03-22
feat: add lockdown mode on multiple login attempts

steveiliop56:
## Summary by CodeRabbit

* **Security Improvements**
  * Added a global automatic lockdown that temporarily blocks login attempts when failures spike; locked status reports remaining seconds until re-open.
  * Enforces a hard cap on stored failed-login records and stops recording beyond that limit to limit exposure.

* **Performance**
  * Clears stored login attempt data when lockdown activates and automatically resumes normal operation after the timeout.

### DIFF
--- a/internal/service/auth_service.go
+++ b/internal/service/auth_service.go
@@ -242,7 +242,7 @@ func (auth *AuthService) RecordLoginAttempt(identifier string, success bool) {
 	defer auth.loginMutex.Unlock()
 
 	if len(auth.loginAttempts) >= MaxLoginAttemptRecords {
-		if auth.lockdown != nil || !auth.lockdown.Active {
+		if auth.lockdown != nil && auth.lockdown.Active {
 			return
 		}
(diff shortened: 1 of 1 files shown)


internal/service/auth_service.go
```diff
@@ -227,6 +241,14 @@ func (auth *AuthService) RecordLoginAttempt(identifier string, success bool) {
 	auth.loginMutex.Lock()
 	defer auth.loginMutex.Unlock()
 
+	if len(auth.loginAttempts) >= MaxLoginAttemptRecords {
+		if auth.lockdown != nil || !auth.lockdown.Active {
```
Rycochet: Are you sure this is correct - early bail if it's *not* active?

steveiliop56: Good catch. Fixed.

steveiliop56:
Yeah we can merge this.

PR #777, opened 2026-04-09
feat(oidc): support for all in-spec attributes and scopes

scottmckendry:
A decent sized refactor that will require us to update the appropriate documentation if it gets merged. I'm happy to do this.

Should be backwards compatible (I think) with the old users config.

5 new test passes:

image

New scopes in authorization:

image

Now, I'm fine if we decide we're not interested in supporting phone number & address. That would make this PR a little bit smaller as a result. Selfishly, I just wanted to turn all of these tests green 🚀 

## Release Notes

* **New Features**
  * Added support for `phone` and `address` OAuth scopes with visual indicators in the authorization flow
  * Extended OIDC userinfo claims to include phone number, address, and profile attributes (given name, family name, middle name, nickname, profile picture, website, gender, birthdate, timezone, locale)
  * Added configuration support for per-user identity attributes


Rycochet:
@coderabbitai review

internal/utils/loaders/loader_file.go
```diff
@@ -32,9 +32,7 @@ func (f *FileLoader) Load(args []string, cmd *cli.Command) (bool, error) {
 
 	log.Warn().Msg("Using experimental file config loader, this feature is experimental and may change or be removed in future releases")
 
-	err = file.Decode(flags[configFileFlag], cmd.Configuration)
-
-	if err != nil {
+	if err = file.Decode(flags[configFileFlag], cmd.Configuration); err != nil {
```
Rycochet: I don't think this helps readability - also doesn't actually change anything, so could revert this file 😅

steveiliop56: I agree tbh, it makes absolutely zero difference but I prefer having the separate if block.

scottmckendry: ah yes, good catch. I knew something like this would fall through the cracks after a few different refactors

internal/config/config.go
```diff
@@ -113,15 +113,44 @@ type ServerConfig struct {
...
+	UsersFile          string                    `description:"Path to the users file." yaml:"usersFile"`
+	SecureCookie       bool                      `description:"Enable secure cookies." yaml:"secureCookie"`
+	SessionExpiry      int                       `description:"Session expiry time in seconds." yaml:"sessionExpiry"`
+	SessionMaxLifetime int                       `description:"Maximum session lifetime in seconds." yaml:"sessionMaxLifetime"`
+	LoginTimeout       int                       `description:"Login timeout in seconds." yaml:"loginTimeout"`
+	LoginMaxRetries    int                       `description:"Maximum login retries." yaml:"loginMaxRetries"`
+	TrustedProxies     []string                  `description:"Comma-separated list of trusted proxy addresses." yaml:"trustedProxies"`
+}
+
+type UserAttributes struct {
```
steveiliop56, line 128: Any chance we could also add the `email_verified` as an attribute? Because right now Tinyauth sets it to true by default.

Alternatively, we could remove all `_verified` attributes (and set them to `true` in the response) since whatever you set in Tinyauth can be considered trusted.

internal/config/config.go
```diff
@@ -254,6 +284,7 @@ type UserContext struct {
 	OAuthName   string
 	OAuthSub    string
 	LdapGroups  string
+	Attributes  UserAttributes
```
steveiliop56, line 286: @Rycochet suggested we remove the attributes from the LDAP users for now and I believe we should do so too.

internal/service/oidc_service.go
```diff
@@ -430,20 +479,16 @@ func (service *OIDCService) generateIDToken(client config.OIDCClientConfig, user
 		return "", err
 	}
 
-	userInfo := service.CompileUserinfo(user, scope)
-
+	// Per OIDC Core §5.4: for code flow, scope-requested claims (profile, email)
+	// belong in the userinfo response only. The id_token carries only the required
+	// JWT claims (iss, aud, sub, iat, exp) plus nonce.
```
steveiliop56: I know, but a lot of apps have made storing claims in the ID token the "standard" so, I added them in Tinyauth too to be compatible with as many providers as possible. Now, the token will probably be massive with all of these claims but what can we do?

internal/service/oidc_service_test.go
steveiliop56, line 1: Can we follow the same testing scheme as the controllers? I believe it's a bit easier to work with and it's also less code ;)

internal/controller/user_controller_test.go
```diff
@@ -353,3 +354,150 @@ func TestUserController(t *testing.T) {
 		require.NoError(t, err)
 	})
 }
+
+func TestUserControllerAttributes(t *testing.T) {
```
steveiliop56: Can we merge this test with the one above? Ideally it should be one test function per controller.

steveiliop56 (changes requested):
Looks really good! Some small feedback and we can merge.

internal/config/config.go
Rycochet: If someone has explicitly set a value then `key_verified=true`, if not set then `""` / `false`?

internal/config/config.go
scottmckendry: I spent way to long staring at this, then going back to my local thinking I was losing it haha.

they way github is rendering the diff makes it look like this is in the ldapuser struct. 

Image

it's actually in usercontext...

internal/service/oidc_service.go
scottmckendry: I've added them back now. It just means there's going to be warnings in some of the scope tests.

>Technically returning unrequested claims does not violate the specifications but it could be a bug in the server and may result in user data being exposed in unintended ways if the relying party did not expect the email to be in the id_token, and then uses the id_token to provide proof of the authentication event to other parties.

frontend/src/pages/authorize-page.tsx
```diff
@@ -61,6 +61,18 @@ const createScopeMap = (t: TFunction<"translation", undefined>): Scope[] => {
...
+      name: t("phoneScopeName"),
+      description: t("phoneScopeDescription"),
+      icon: <Phone {...scopeMapIconProps} />,
+    },
+    {
+      id: "address",
+      name: t("addressScopeName"),
+      description: t("addressScopeDescription"),
+      icon: <MapPin {...scopeMapIconProps} />,
+    },
```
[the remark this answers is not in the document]
scottmckendry: not in scope for this pr, right now. I only speak english :)

internal/config/config.go
```diff
@@ -113,15 +113,44 @@ type ServerConfig struct {
...
+	Website             string       `description:"URL of the user's website." yaml:"website"`
+	Email               string       `description:"Email address of the user." yaml:"email"`
+	Gender              string       `description:"Gender of the user." yaml:"gender"`
+	Birthdate           string       `description:"Birthdate of the user (YYYY-MM-DD)." yaml:"birthdate"`
+	Zoneinfo            string       `description:"Time zone of the user (e.g. Europe/Athens)." yaml:"zoneinfo"`
+	Locale              string       `description:"Locale of the user (e.g. en-US)." yaml:"locale"`
+	PhoneNumber         string       `description:"Phone number of the user." yaml:"phoneNumber"`
+	PhoneNumberVerified bool         `description:"Whether the phone number has been verified." yaml:"phoneNumberVerified"`
+	Address             AddressClaim `description:"Address of the user." yaml:"address"`
+}
```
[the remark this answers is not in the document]
scottmckendry: tinyauth is self-hosted, so PII isn't really a problem IMO. can revisit later if this becomes a problem.

scottmckendry:
@coderabbitai review

internal/assets/migrations/000008_oidc_userinfo_profile.down.sql
```diff
@@ -0,0 +1,14 @@
...
+ALTER TABLE "oidc_userinfo" DROP COLUMN "nickname";
+ALTER TABLE "oidc_userinfo" DROP COLUMN "profile";
+ALTER TABLE "oidc_userinfo" DROP COLUMN "picture";
+ALTER TABLE "oidc_userinfo" DROP COLUMN "website";
+ALTER TABLE "oidc_userinfo" DROP COLUMN "gender";
+ALTER TABLE "oidc_userinfo" DROP COLUMN "birthdate";
+ALTER TABLE "oidc_userinfo" DROP COLUMN "zoneinfo";
+ALTER TABLE "oidc_userinfo" DROP COLUMN "locale";
+ALTER TABLE "oidc_userinfo" DROP COLUMN "phone_number";
+ALTER TABLE "oidc_userinfo" DROP COLUMN "phone_number_verified";
```
steveiliop56: Let's remove this line since we don't have a config option for the phone number verified anymore.

internal/controller/user_controller.go
```diff
@@ -144,6 +161,15 @@ func (controller *UserController) loginHandler(c *gin.Context) {
...
 
+	if userSearch.Type == "local" && localUser != nil {
+		if localUser.Attributes.Name != "" {
+			sessionCookie.Name = localUser.Attributes.Name
+		}
+		if localUser.Attributes.Email != "" {
+			sessionCookie.Email = localUser.Attributes.Email
+		}
+	}
+
```
steveiliop56, line 172: We probably need automatic generation here too. Probably logic should be we generate by default and if it's local we check for the attributes.

steveiliop56 (changes requested):
Final comments. Can you please confirm that attributes are set in the `remote-` headers in the authentication response?

internal/controller/user_controller.go
scottmckendry: Isn't this what we're already doing a few lines up from this?

```go
	sessionCookie := repository.Session{
		Username: req.Username,
		Name:     utils.Capitalize(req.Username),
		Email:    utils.CompileUserEmail(req.Username, controller.config.CookieDomain),
		Provider: "local",
	}
```

Sorry if I've misunderstood 🙂

steveiliop56: Oh didn't notice it, yeah it's fine.

steveiliop56:
@scottmckendry would you like to solve conflicts or should I? It's just the go module rename.

scottmckendry:
@steveiliop56 - good to go. I've left in my forked branch because I'm not sure how to change upstream without having to re-open the PR.. and this one has quite a bit of history 😁

steveiliop56 (approved):

scottmckendry:
@dosubot update docs

PR #785, opened 2026-04-11
feat: support unsigned oidc request objects

scottmckendry:
advertises `request_parameter_supported` and `request_object_signing_alg_values_supported` in discovery

decodes jwt in frontend to merge claims

now passes the "oidcc-unsigned-request-object-supported-correctly-or-rejected-as-unsupported" conformance test:

image

## Summary by CodeRabbit

* **New Features**
  * Added support for OpenID Connect "request" parameter processing so request objects can supply/override login parameters.
  * OpenID configuration endpoint now advertises support for the request parameter and the allowed request object signing algorithm "none".
* **Tests**
  * Updated configuration endpoint tests to cover the new advertised fields.

### DIFF
--- a/frontend/src/lib/hooks/oidc.ts
+++ b/frontend/src/lib/hooks/oidc.ts
@@ -11,6 +11,33 @@ export const oidcParamsSchema = z.object({
   code_challenge_method: z.string().optional(),
 });
 
+function b64urlDecode(s: string): string {
+  const base64 = s.replace(/-/g, "+").replace(/_/g, "/");
+  return atob(base64.padEnd(base64.length + ((4 - (base64.length % 4)) % 4), "="));
+}
+
+function decodeRequestObject(jwt: string): Record<string, string> {
+  try {
+    // Must have exactly 3 parts: header, payload, signature
+    const parts = jwt.split(".");
+    if (parts.length !== 3) return {};
+
+    // Header must specify "alg": "none" and signature must be empty string
+    const header = JSON.parse(b64urlDecode(parts[0]));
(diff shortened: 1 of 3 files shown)


frontend/src/lib/hooks/oidc.ts
```diff
@@ -11,6 +11,33 @@ export const oidcParamsSchema = z.object({
...
+    if (!payload || typeof payload !== "object" || Array.isArray(payload)) return {};
+    const result: Record<string, string> = {};
+    for (const [k, v] of Object.entries(payload)) {
+      if (typeof v === "string") result[k] = v;
+    }
+    return result;
+  } catch {
+    return {};
+  }
+}
```
steveiliop56, line 39: Perhaps we should use a JWT library to handle the decoding? Is it worth the additional 50kb? https://bundlephobia.com/package/jsonwebtoken@9.0.3

scottmckendry: I don't think this library will help much here, pretty sure that's for node.js server stuff, not client side. 

We could add jose here (I've used in couple next.js projects before), which is 60kb?

However, that would only be useful if/when we add support for signed jwts - other than that it would only handle the base64 and header parsing. It looks like jose actually rejects alg=none by default, so we'd need to keep the majority of this code around.

steveiliop56: Makes sense, since the spec says MAY we can ignore signed ones and JWEs for now.

steveiliop56 (approved):

PR #786, opened 2026-04-11
fix: revoke access token on duplicate auth code user

steveiliop56:
image

## Summary by CodeRabbit

* **New Features**
  * Access tokens are now invalidated if an authorization code is reused; reusing a code returns an "invalid_grant" error and previously issued tokens become unusable (subsequent API calls fail).

* **Tests**
  * Added test coverage confirming tokens are invalidated on double use and that appropriate errors are returned.


internal/assets/migrations/000008_oidc_coder_user.up.sql
scottmckendry, line 1: you're probably already aware - but this pr and #777 both introduce migration 08 so we'll need to handle this conflict in on or the other depending on which is merged first.

scottmckendry (approved):

internal/repository/oidc_queries.sql.go
```diff
@@ -219,6 +223,7 @@ func (q *Queries) DeleteExpiredOidcTokens(ctx context.Context, arg DeleteExpired
 			&i.Sub,
 			&i.AccessTokenHash,
 			&i.RefreshTokenHash,
+			&i.CodeHash,
```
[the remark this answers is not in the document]
steveiliop56: But didn't we add `DEFAULT ""` which won't be null?

internal/assets/migrations/000008_oidc_coder_user.up.sql
steveiliop56: @scottmckendry I know but since this PR is basically a security fix can you change your migration to 09?

Rycochet (approved):

PR #800, opened 2026-04-16
fix: preserve OIDC params during login → TOTP redirect

frankhommers:
## Bug

When a user with TOTP enabled tries to log in to an app via tinyauth's OIDC provider, the first login attempt always fails with a "State does not match" error on the client app (e.g. Vikunja, Calibre-Web). The second attempt works fine.

This affects **all OIDC client apps** connected to tinyauth when the user has TOTP enabled and is not yet logged in to tinyauth.

## Root Cause

The OIDC authorization flow goes through these pages:

1. `/authorize?scope=...&client_id=...&state=...` → user not logged in
2. React Router `<Navigate>` to `/login?scope=...&client_id=...&state=...`
3. User enters password → server responds with `{ totpPending: true }`
4. `onSuccess` callback redirects to `/totp?...`

The problem is in step 4. The `onSuccess` callback uses `oidcParams` from a closure that was captured at render time via `useLocation().search`. However, when the authorize page redirects to the login page via React Router's `<Navigate replace />`, the `useLocation()` hook can return stale search params that don't include the OIDC parameters.

As a result, `oidcParams.isOidc` is `false` in the callback, and the redirect falls through to the else-branch which only passes `redirect_uri`:

```
// What happens (broken):
/totp?redirect_uri=https://app.example.com/callback

// What should happen (fixed):
/totp?scope=openid+email+profile&response_type=code&client_id=...&redirect_uri=...&state=...
```

After TOTP verification, the TOTP page sees `oidcParams.isOidc = false` (missing params) and redirects to `/continue` instead of `/authorize?...`. The user ends up back at the app without a valid OIDC code/state, causing the "State does not match" error.

The second attempt works because the user is already logged in at tinyauth — the authorize page shows directly (no login/TOTP needed), so no params are lost.

## Fix

Re-parse OIDC params from `window.location.search` inside the `onSuccess` callback instead of relying on the potentially stale closure from `useLocation()`. `window.location.search` is always current and reliably contains the OIDC parameters.

To make the parsing function callable outside of React components, the existing `useOIDCParams` hook logic is extracted into a plain `parseOIDCParams` function. `useOIDCParams` becomes a thin wrapper around it (no breaking changes).

## Verified

Tested end-to-end with Playwright against a live tinyauth + Vikunja setup:

- Clean browser state (no cookies/storage)
- Vikunja → tinyauth login → TOTP → authorize → Vikunja callback
- All OIDC params (`scope`, `client_id`, `state`, `redirect_uri`) preserved through the entire flow
- User successfully logged in on first attempt

## Summary by CodeRabbit

* **Refactor**
  * Restructured OIDC parameter parsing logic for improved code reusability and maintainability.
  * Enhanced login flow to ensure OIDC parameters are freshly evaluated during authentication completion, improving reliability of OAuth/OIDC integrations.

### DIFF
--- a/frontend/src/lib/hooks/oidc.ts
+++ b/frontend/src/lib/hooks/oidc.ts
@@ -38,14 +38,14 @@ function decodeRequestObject(jwt: string): Record<string, string> {
   }
 }
 
-export const useOIDCParams = (
-  params: URLSearchParams,
-): {
+export type OIDCParamsResult = {
   values: z.infer<typeof oidcParamsSchema>;
   issues: string[];
   isOidc: boolean;
   compiled: string;
-} => {
+};
+
+export const parseOIDCParams = (params: URLSearchParams): OIDCParamsResult => {
   const obj = Object.fromEntries(params.entries());
 
   // RFC 9101 / OIDC Core 6.1: if `request` param present, decode JWT payload
@@ -73,4 +73,8 @@ export const useOIDCParams = (
     isOidc: false,
     compiled: "",
   };
+};
+
+export const useOIDCParams = (params: URLSearchParams): OIDCParamsResult => {
+  return parseOIDCParams(params);
 };
--- a/frontend/src/pages/login-page.tsx
+++ b/frontend/src/pages/login-page.tsx
@@ -18,7 +18,7 @@ import { OAuthButton } from "@/components/ui/oauth-button";
 import { SeperatorWithChildren } from "@/components/ui/separator";
 import { useAppContext } from "@/context/app-context";
 import { useUserContext } from "@/context/user-context";
-import { useOIDCParams } from "@/lib/hooks/oidc";
+import { useOIDCParams, parseOIDCParams } from "@/lib/hooks/oidc";
 import { LoginSchema } from "@/schemas/login-schema";
 import { useMutation } from "@tanstack/react-query";
 import axios, { AxiosError } from "axios";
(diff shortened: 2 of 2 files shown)


scottmckendry:
I'm not able to reproduce the issue on the latest alpha release with Grafana's generic oauth. Observed behaviour:

1. Open grafana, sign-in with tinyauth
2. enter username/password
3. enter totp
4. consent to requested scopes on authorize page
5. successfully redirected back to grafana

I've not tested on the stable release, but we've made a few other changes in this area recently, so perhaps this fix might not be necessary.

Can you test with the latest alpha (5.0.7) and confirm if you still get the problem?

steveiliop56:
@scottmckendry @frankhommers I tested against Vikunja `v2.3.0` and Tinyauth `v5.0.6` and I cannot reproduce this issue. My configuration Tinyauth side was:

```yaml
services:
  tinyauth:
    environment:
      - TINYAUTH_OIDC_CLIENTS_VIKUNJA_CLIENTID=some-client-id
      - TINYAUTH_OIDC_CLIENTS_VIKUNJA_CLIENTSECRET=some-client-secret
      - TINYAUTH_OIDC_CLIENTS_VIKUNJA_TRUSTEDREDIRECTURIS=http://localhost:3456/auth/openid/tinyauth
      - TINYAUTH_OIDC_CLIENTS_VIKUNJA_NAME=Vikunja
```

My Vikunja compose was:

```yaml
services:
  vikunja:
    image: vikunja/vikunja:2.3.0
    ports:
      - 3456:3456
    environment:
      VIKUNJA_SERVICE_PUBLICURL: http://localhost:3456/
      VIKUNJA_SERVICE_JWTSECRET: some-secret
      VIKUNJA_DATABASE_PATH: /db/vikunja.db
    volumes:
      - ./config.yml:/etc/vikunja/config.yml
      - ./files:/app/vikunja/files
      - ./db:/db
      - /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt:ro
    restart: unless-stopped
```

My Vikunja config file was:

```yaml
auth:
  openid:
    enabled: true
    providers:
      tinyauth:
        name: Tinyauth
        authurl: https://tinyauth.example.com
        clientid: some-client-id
        clientsecret: some-client-secret

  local:
    enabled: false
```

Everything worked first try with TOTP. What I did observe is that authentication initially failed due to a misconfiguration and Tinyauth redirected to `http://localhost:3456/auth/openid/tinyauth?code=cBVPWetYvnTcZftopCk5ojm1n1adFzec&state=kanx3km658` (as expected) and I got the state mismatch error.

> closure that was captured at render time

I don't really understand what this means. The `useLocation` hook does not return stale parameters, I don't know from where you got this information. Tinyauth follows correctly the usage of this hook based on the [example](https://reactrouter.com/api/hooks/useLocation).

Since neither me nor Scott can replicate this issue, I am closing this pull request. If some stale hook output is indeed the case, please try to share the output of `window.location` and `useLocation().search` which should not match.

frankhommers:
I don't know what's wrong then. It works with this fix, and does NOT without at my side.

I'll reinspect my configuration then.

Edit: After upgrading to 5.0.7 this fix wasn't necessary anymore and it works without the fix in this PR.

PR #831, opened 2026-04-30
refactor(db): use new store interface

scottmckendry:
lays the groundwork for adding more storage drivers in the future (postgres, redis) with slightly lower effort and easier maintainability in the future.

moves sqlite specific queries, migrations and generated go code into subdirs.

## Summary by CodeRabbit

* **New Features**
  * Configurable database driver (defaults to SQLite); pluggable storage with SQLite and in-memory backends.
  * Migration assets scoped to the SQLite migration set.

* **Chores**
  * CI now enforces generated repository code is up to date.
  * Dev compose updated to enable HTTPS/TLS for local Traefik.
  * Updated Go tooling and changed code-generation entrypoint.

* **Tests**
  * Added comprehensive in-memory repository tests; controller and middleware tests switched to use the in-memory store.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyauthapp/tinyauth/pull/831?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


scottmckendry:
The code gen added in the most recent commit was largely written by an LLM. I have vetted to the best of my ability, but this is a new concept to me - hence this disclaimer.

that out of the way, the generate code satisfies the store interface and correctly wraps all of the sqlc-generated code.

internal/controller/oidc_controller_test.go
```diff
@@ -848,14 +847,10 @@ func TestOIDCController(t *testing.T) {
...
+	store := memory.New()
 
-	db, err := app.SetupDatabase(path.Join(tempDir, "tinyauth.db"))
-	require.NoError(t, err)
-
-	queries := repository.New(db)
-	oidcService := service.NewOIDCService(oidcServiceCfg, queries)
-	err = oidcService.Init()
+	oidcService := service.NewOIDCService(oidcServiceCfg, store)
+	err := oidcService.Init()
```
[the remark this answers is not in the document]
scottmckendry: the previous implementation was also a shared db.

scottmckendry:
@coderabbitai review

cmd/gen/sqlc-wrapper/store.tmpl
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func mapErr(err error) error {
+	for _, e := range errMap {
+		if errors.Is(err, e.from) {
+			return e.to
+		}
+	}
+	return err
+}
```
steveiliop56: Could we use a map instead? E.g.

```go
var errorMap = map[error]error{
  sql.ErrNoRows: repository.ErrNotFound,
}
```

So then we can just do `errorMap[error]`

internal/bootstrap/app_bootstrap.go
```diff
@@ -43,7 +43,7 @@ type BootstrapApp struct {
 	log      *logger.Logger
 	ctx      context.Context
 	cancel   context.CancelFunc
-	queries  *repository.Queries
+	queries  repository.Store
```
steveiliop56, line 47: Why not use a pointer? I believe it's a better idea to point to the store rather than own it.

internal/repository/sqlite/store.go
```diff
@@ -0,0 +1,224 @@
...
+	return repository.OidcCode(v)
+}
+func oidcTokenToRepo(v OidcToken) repository.OidcToken {
+	return repository.OidcToken(v)
+}
+func oidcUserinfoToRepo(v OidcUserinfo) repository.OidcUserinfo {
+	return repository.OidcUserinfo(v)
+}
+func sessionToRepo(v Session) repository.Session {
+	return repository.Session(v)
```
steveiliop56: Is there really a need for a helper function? I mean since we already generate the code we might as well just use `repository.Session` directly.

Makefile
```diff
@@ -84,4 +84,4 @@ sql:
 
 # Go gen
 generate:
-	go run ./gen
+	go generate ./internal/repository/...
```
steveiliop56, line 88: Please keep the `./gen` too since it generates the `.env.example` and documentation reference.

.github/workflows/ci.yml
```diff
@@ -26,6 +26,12 @@ jobs:
       - name: Go dependencies
         run: go mod download
 
+      - name: Check codegen is up to date
+        run: |
+          go generate ./internal/repository/...
+          git diff --exit-code -- internal/repository/
+          git status --porcelain -- internal/repository/ | grep -q . && echo "untracked files in internal/repository/" && exit 1 || true
+
```
steveiliop56, line 42: Should the CI be responsible for this? It doesn't make sense to me to try to generate code in the CI because then why not generate SQLC code too?

cmd/gen/sqlc-wrapper/sqlc_wrapper.go
```diff
@@ -0,0 +1,527 @@
...
+	if err != nil {
+		return nil, mapErr(err)
+	}
+	out := make([]{{.RepoType}}, len(rows))
+	for i, row := range rows {
+		out[i] = {{.Converter}}(row)
+	}
+	return out, nil
+{{end}}`),
+)
```
steveiliop56: Maybe it would be better to have this as a template file too?

steveiliop56 (changes requested):

internal/bootstrap/app_bootstrap.go
scottmckendry: this is an interface now, not a struct. It's generally not a good idea to use pointers to interfaces (I'm not sure it even works).

.github/workflows/ci.yml
scottmckendry: I wanted a way to make sure PRs weren't manually editing generated files. 

This is more of test than actually generating the result.

Originally i had planned on having sqlc in here too, but it would mean having a fixed version of the sqlc binary used to generate. otherwise you get false positives on the version comment at the top of the gen'd files.

steveiliop56: Oh I like this, I misunderstood. Well maybe we should bundle everything into a small script? There is nothing stopping us from adding SQLC to be honest. We can just read the files, get the version and check if anything meaningful has changed with the correct binary version. I wonder, could we make this into an action ; )

internal/bootstrap/app_bootstrap.go
steveiliop56: Makes sense.

gen/sqlc-wrapper/store.tmpl
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func mapErr(err error) error {
+	for from, to := range errorMap {
+		if errors.Is(err, from) {
+			return to
+		}
+	}
+	return err
+}
```
steveiliop56, line 33: Is it possible to do:

```go
func mapErr(err error) error {
	merr, ok := errorMap[err]
    if ok { return merr }
    return err
}
```

Or are we limited by the "dynamic" nature of the errors (each time `errors.New` returns a slightly different error)?

internal/bootstrap/app_bootstrap.go
```diff
@@ -173,12 +174,13 @@ func (app *BootstrapApp) Setup() error {
 	defer func() {
 		app.cancel()
 		app.wg.Wait()
-		app.db.Close()
+		if app.db != nil {
+			app.db.Close()
+		}
```
steveiliop56, line 179: Why would the db be nil here? This defer is defined after the db is initialized.

internal/repository/memory/memory_test.go
steveiliop56, line 1: Would it make sense to test the memory driver in the same why as we test services and controllers?

gen/sqlc-wrapper/store.tmpl
scottmckendry: direct lookup won't work for this implementation, unfortunately.

internal/bootstrap/app_bootstrap.go
scottmckendry: it's nil if it's a `memory` driver - this may panic without the check. Something coderabbit picked up in one of it's earlier comments

internal/repository/memory/memory_test.go
scottmckendry: done

internal/repository/memory/memory_test.go
```diff
@@ -0,0 +1,427 @@
...
+	})
+	require.NoError(t, err)
+
+	deleted, err := s.DeleteExpiredOidcTokens(ctx, repository.DeleteExpiredOidcTokensParams{
+		TokenExpiresAt:        50,
+		RefreshTokenExpiresAt: 50,
+	})
+	require.NoError(t, err)
+	assert.Len(t, deleted, 2)
+
```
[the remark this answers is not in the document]
scottmckendry: sqlite should be the source of truth here. I've updated the memory driver to match now

steveiliop56 (approved):
LGTM

scottmckendry:
@dosubot update docs

Issue #717, opened 2026-03-16
[FEATURE] Server Socketpath Permissions


MagnusFriendz:
**Is your feature request related to a problem? Please describe.**
I have Tinyauth running under a different user than Caddy. With the default configuration Tinyauth is owning its socket and gives it 755 permissions. Caddy cannot use it because it only can read but not write.

**Describe the solution you'd like**
- Something like a VAR/flag TINYAUTH_SERVER_SOCKETPERMISSION=777 (or at least 775)
- or maybe another solution to give Caddy r/w access 

**Describe alternatives you've considered**
- chmod the socket manually on every restart which is not practicable in production
- run Tinyauth under the Caddy User or vice versa which is insecure, I do not want the webserver to be able to read all the  sensitive Tinyauth configuration 
- even if I do it in reverse, adding Caddy to the Tinyauth group, Caddy has still only read access
- an auxiliary group (with caddy and tinyauth as members) is useless as well because the permissions for groups are read-only 

**Additional context**
I consider a unix socket with permissions 777 residing in a restricted folder more secure than using a tcp socket without tls encryption to which everyone on the network potentially listens to

I'd be happy if you could consider this enhancement. 

Have a good day,
Magnus.

steveiliop56:
I don't believe a configuration option is necessary for this.

The 775 downgrade from 755 doesn't sound all that bad. I can include it in the next patch release.

MagnusFriendz:
Simple. Perfekt. 

Thank you.

steveiliop56:
Seems like you cannot change the socket permissions easily.

> Caddy cannot use it because it only can read but not write.

Caddy should need to write to this file. Sockets are not meant for writing. It should just be able to read it.

PR #833, opened 2026-05-02
bootstrap: set UNIX socket permissions to 770

ungeskriptet:
This grants other processes access to the socket, provided they are in the correct group.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced UNIX socket permission configuration when the application is configured to run on a UNIX socket, with improved error handling for permission operations.

### DIFF
--- a/internal/bootstrap/app_bootstrap.go
+++ b/internal/bootstrap/app_bootstrap.go
@@ -215,6 +215,12 @@ func (app *BootstrapApp) Setup() error {
 		}
 
 		tlog.App.Info().Msgf("Starting server on unix socket %s", app.config.Server.SocketPath)
+		go func() {
+			// Ensure processes running as a different user can access the socket.
+			if err := os.Chmod(app.config.Server.SocketPath, 0770); err != nil {
+				tlog.App.Fatal().Err(err).Msg("Failed to update UNIX socket permissions")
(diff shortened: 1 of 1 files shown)


internal/bootstrap/app_bootstrap.go
```diff
@@ -215,6 +215,12 @@ func (app *BootstrapApp) Setup() error {
 		}
 
 		tlog.App.Info().Msgf("Starting server on unix socket %s", app.config.Server.SocketPath)
+		go func() {
```
scottmckendry, line 218: I think umask would be the more correct approach here, rather than chmod.

scottmckendry (changes requested):
Maybe I'm out of the loop, but could you also describe some use cases for this change?

internal/bootstrap/app_bootstrap.go
steveiliop56: I agree, AFAIK unix sockets don't really have permissions by themselves no? Don't they follow the permissions of the directory they are in? At least that's what I remember reading in stackoverflow.

steveiliop56:
@scottmckendry I am assuming this PR solves https://github.com/tinyauthapp/tinyauth/issues/717

PR #895, opened 2026-05-24
feat: Add support for custom username overrides

mathieu-lemay:
Add a new option to manually define the usernames for selected email addresses. This is mostly useful for OAuth providers that don't return `preferred_username`.

I'm opening this as a draft since it's a quick and dirty fix. If this is something you're interested in, I will take the time to add some validations (like email / username is not empty) and some tests.


steveiliop56:
@mathieu-lemay thank you for the pull request.

I am a bit skeptical about it. Why add another file, another field to manage and another mapping while we already have user attributes? Can't the same outcome be achieved by adding a new `PreferredUsername` field in the `UserAttributes` struct? Granted, it will require the user attributes to also mess with OAuth users but I believe it's a better idea because what if we also want a custom name for OAuth users in the future? Another file?

mathieu-lemay:
@steveiliop56 Thanks for the reply, this is exactly the sort of feedback I wanted! I didn't think of using UserAttributes because I haven't used that feature in my workflow with tinyauth. I'll look into it and get back to you

mathieu-lemay:
@steveiliop56 I've updated it to use the existing user attributes. I had to implement some slugify like [function](https://github.com/tinyauthapp/tinyauth/pull/895/changes#diff-c450b2f3fa0c3d8e3395b93a3cd6921efa302d66af05d1b7087062d167b7f991R417-R422) to support having emails in env var names. I'm not sure it's worth pulling in a new dependency for that, this should be enough.

I also added an "override" for the provider overrides, since it's possible to have groups for an overriden provider now.

Anyway, tell me what you think. To be frank, I don't even need this anymore since I switched from Google to PocketId, but it could still be interesting for others.

Issue #817, opened 2026-04-24
[FEATURE] add OAUTH_WHITELISTFILE support (file-based alternative to OAUTH_WHITELIST env var)


falkheiland:
**Is your feature request related to a problem? Please describe.**
Managing longer whitelists as a comma-separated string in a compose file is cumbersome and error-prone (see also #649, where long whitelist strings caused silent truncation/rejection issues).

**Describe the solution you'd like**
Add a `FILE` equivalent for `OAUTH_WHITELIST` (and ideally `OAUTH_BLACKLIST`) following the same pattern already established for `USERS` / `USERSFILE`:

- `TINYAUTH_APPS_name_OAUTH_WHITELISTFILE` pointing to a file containing one email address per line
- Consistent with the existing `--auth.usersfile` / `TINYAUTH_AUTH_USERSFILE` implementation

PR #898, opened 2026-05-24
feat: add support for app-level oauth whitelist file

djedditt:
Follow-up to #817.

Added `TINYAUTH_APPS_name_OAUTH_WHITELISTFILE` support for loading app-specific OAuth whitelist entries from a file, merged with the existing inline `TINYAUTH_APPS_name_OAUTH_WHITELIST` config.

(Also corrected the app OAuth whitelist description from "groups" to "emails")

## Summary by CodeRabbit

* **New Features**
  * Added support for loading OAuth email whitelists from external configuration files alongside inline settings
  * System now merges email whitelists from both inline and file-based sources

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyauthapp/tinyauth/pull/898?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -62,8 +62,10 @@ TINYAUTH_APPS_name_CONFIG_DOMAIN=
 TINYAUTH_APPS_name_USERS_ALLOW=
 # Comma-separated list of blocked users.
 TINYAUTH_APPS_name_USERS_BLOCK=
-# Comma-separated list of allowed OAuth groups.
+# Comma-separated list of allowed OAuth emails.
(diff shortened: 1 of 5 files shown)


steveiliop56:
@djedditt maybe we should move the whitelist fetching directly in the ACL rule?

https://github.com/tinyauthapp/tinyauth/blob/49105ce5ff066f0dd642af2fa7d6a453fa623258/internal/service/access_controls_rules.go#L44

It's not used anywhere else so we can just check it directly there without needing to add overhead to the ACL service.

PR #913, opened 2026-05-31
refactor: rework oidc session storage

steveiliop56:
## Summary by CodeRabbit

* **New Features**
  * Added PostgreSQL as a documented database-driver option in the example env.

* **Refactor**
  * Consolidated OpenID Connect storage to a single session model: tokens are session-based, userinfo is stored with sessions, code handling moved to cached/ session flows, and periodic cleanup simplified for more consistent auth state.


sql/sqlite/oidc_queries.sql
```diff
@@ -1,133 +1,48 @@
...
+UPDATE "oidc_sessions" SET
+    "access_token_hash" = ?,
+    "refresh_token_hash" = ?,
+    "scope" = ?,
+    "client_id" = ?,
+    "token_expires_at" = ?,
+    "refresh_token_expires_at" = ?,
+    "nonce" = ?,
+    "userinfo_json" = ?
+WHERE "sub" = ?
```
scottmckendry, line 47: Correct me if I'm wrong, but shouldn't this use the `refresh_token_hash` instead of the `sub` to prevent replay?

internal/service/oidc_service.go
```diff
@@ -426,36 +464,24 @@ func (service *OIDCService) ValidateGrantType(grantType string) error {
 	return nil
 }
 
-func (service *OIDCService) GetCodeEntry(c *gin.Context, codeHash string, clientId string) (repository.OidcCode, error) {
-	oidcCode, err := service.queries.GetOidcCode(c, codeHash)
+func (service *OIDCService) GetCodeEntry(codeHash string, clientId string) (*AuthorizeCodeEntry, bool) {
```
scottmckendry, line 467: should this function use the cache services `WithLock` to make the get/delete atomic?

There's possibly a race here that allows the same code to be used twice with the right timing. Not sure if that's a vuln or not.

scottmckendry (dismissed):
Two minor comments - not sure if they need to be resolved to merge so will approve now. 

Let me know if I need to re-approve.

internal/service/oidc_service.go
steveiliop56: Good catch. Unlikely/maybe impossible to happen since we also have the code reuse but it's a good idea to have since we already have the `WithLock`.

sql/sqlite/oidc_queries.sql
steveiliop56: Not really. I used the `sub` as a generic lookup for updating not tied to the refresh logic. How would a replay work?

scottmckendry: If it's separate from refresh then should be fine. I did check to see where this was being used.

scottmckendry (approved):

PR #929, opened 2026-06-08
Add LDAP BindPasswordFile

### LINKED ISSUE #927: [FEATURE] Add support for docker secrets for ldap - TINYAUTH_LDAP_BINDPASSWORDFILE
### Is your feature request related to a problem? Please describe.

Currently you have to pass a plain text password as an env variable - `TINYAUTH_LDAP_BINDPASSWORD` - this means the password is visible to anything with docker socket access.

LLDAP itself supports this already with `LLDAP_LDAP_USER_PASS_FILE`

### Describe the solution you'd like.

Add support to pass a file as an alternative - `TINYAUTH_LDAP_BINDPASSWORDFILE`

- [x] I confirm this request was written by me and not generated by an LLM or AI assistant.

Rycochet:
Fixes #927

## Summary by CodeRabbit

* **New Features**
  * LDAP bind password can now be supplied via a file path as an alternative to embedding the password in configuration.

* **Bug Fixes / Behavior**
  * When a file path is provided, the password is loaded into memory before binding and the file-based reference is cleared so subsequent operations use the in-memory secret.

### DIFF
--- a/internal/service/ldap_service.go
+++ b/internal/service/ldap_service.go
@@ -33,6 +33,10 @@ func NewLdapService(
 		return nil, nil
 	}
 
+	secret := utils.GetSecret(config.LDAP.BindPassword, config.LDAP.BindPasswordFile)
+	config.LDAP.BindPassword = secret
+	config.LDAP.BindPasswordFile = ""
+
 	ldap := &LdapService{
 		log:    log,
 		config: config,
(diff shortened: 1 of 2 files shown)


internal/service/ldap_service.go
```diff
@@ -212,6 +213,9 @@ func (ldap *LdapService) BindService(rebind bool) error {
 	if ldap.cert != nil {
 		return ldap.conn.ExternalBind()
 	}
+	secret := utils.GetSecret(ldap.config.LDAP.BindPassword, ldap.config.LDAP.BindPasswordFile)
+	ldap.config.LDAP.BindPassword = secret
+	ldap.config.LDAP.BindPasswordFile = ""
```
steveiliop56: @Rycochet this is not an ideal approach. This will re-read the password every time the bind is called (which is called on every authentication). I would recommend moving this to the initialization function instead.

steveiliop56 (changes requested):

steveiliop56 (approved):

PR #948, opened 2026-06-19
feat: support for the prompt parameter in the oidc flow

steveiliop56:
## Summary by CodeRabbit

* **New Features**
  * Added end-to-end support for the OpenID Connect `prompt` parameter (`none`/`login`), including URL handling and backend enforcement.
  * ID tokens now include `auth_time`.

* **Improvements**
  * `prompt=none` now returns a standardized “login required” error when silent sign-in isn’t possible.
  * Requests that combine `prompt=none` with other prompt values are rejected as invalid.
  * Updated login/authorize page behavior for auto-authorization, redirects, and disabling actions during auto-authorization.

### DIFF
--- a/frontend/src/pages/login-page.tsx
+++ b/frontend/src/pages/login-page.tsx
@@ -65,7 +65,7 @@ export const LoginPage = () => {
   const screenParams = useScreenParams(searchParams);
   const compiledParams = recompileScreenParams({
     ...screenParams,
-    oidc_login: false,
+    oidc_prompt: undefined,
   });
   const loginForUrl = useLoginFor({
     login_for: screenParams.login_for,
@@ -199,7 +199,7 @@ export const LoginPage = () => {
     };
   }, [redirectTimer, redirectButtonTimer]);
 
-  if (auth.authenticated && !screenParams.oidc_login) {
+  if (auth.authenticated && screenParams.oidc_prompt !== "login") {
     return <Navigate to={loginForUrl} replace />;
   }
 
--- a/frontend/src/pages/login-page.tsx
+++ b/frontend/src/pages/login-page.tsx
@@ -63,7 +63,10 @@ export const LoginPage = () => {
 
   const searchParams = new URLSearchParams(search);
   const screenParams = useScreenParams(searchParams);
-  const compiledParams = recompileScreenParams(screenParams);
+  const compiledParams = recompileScreenParams({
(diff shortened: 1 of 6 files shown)


frontend/src/pages/login-page.tsx
```diff
@@ -63,7 +63,10 @@ export const LoginPage = () => {
 
   const searchParams = new URLSearchParams(search);
   const screenParams = useScreenParams(searchParams);
-  const compiledParams = recompileScreenParams(screenParams);
+  const compiledParams = recompileScreenParams({
+    ...screenParams,
+    oidc_login: false,
+  });
```
Rycochet, line 69: Is this deliberately trying to set `oidc_login: false` every time, or should it be default = false and overridable?

steveiliop56: Yes we need to set it every time. This is because:

- First scenario, you are not logged in. So you log in and then there is no reason to login again since you just logged in so we remove the `prompt=login`.
- Second scenario, you are logged in and you get prompted to log back in. After you login, if we don't unset the field then the entire flow will repeat (login screen -> authorize screen -> login screen).

Rycochet (dismissed):

internal/service/oidc_service.go
```diff
@@ -660,7 +674,7 @@ func (service *OIDCService) RefreshAccessToken(ctx context.Context, refreshToken
 
 	idToken, err := service.generateIDToken(model.OIDCClientConfig{
 		ClientID: entry.ClientID,
-	}, userInfo, entry.Scope, entry.Nonce)
+	}, userInfo, entry.Scope, entry.Nonce, 0) // auth_time is not available during refresh, so we set it to 0
```
scottmckendry: Is `0` correct here, or should we omit?

if we make the shape:
```go
func (service *OIDCService) generateIDToken(client model.OIDCClientConfig, user UserinfoResponse, scope string, nonce string, authTime *int64) (string, error) 
```

We could pass `nil` here instead.

scottmckendry (dismissed):
very small nit that may be more correct for OIDC spec. Otherwise LGTM 🚀

internal/service/oidc_service.go
steveiliop56: Done.

scottmckendry (approved):

PR #962, opened 2026-07-02
refactor: use some colors in CLI output

steveiliop56:
## Summary by CodeRabbit

* **New Features**
  * Added a `help` subcommand and improved `version` output styling.
  * Updated `config`, `create`, `generate`, `verify`, and OIDC client generation to show YAML-focused, styled console output; OIDC client flags renamed to `--oidc-clients-<name>-*`.
* **Bug Fixes**
  * Stronger input validation with clearer help output when required fields are missing.
  * Improved TOTP instructions/verification messaging and more consistent Docker escaping in rendered values.
* **Chores**
  * Updated configuration YAML tags to be optional (`omitempty`) in generated YAML (except a Tailscale-related field).


cmd/tinyauth/config.go
```diff
@@ -1,25 +1,46 @@
...
+
+			for l := range strings.SplitSeq(string(yout), "\n") {
+				if l == "" {
+					continue
+				}
+				lp := strings.SplitN(l, ":", 2)
+				buf.WriteString(redStyle.Render(lp[0]))
+				buf.WriteString(grayStyle.Render(":"))
+				if len(lp) == 2 {
+					buf.WriteString("")
```
Rycochet: Why are you writing an empty string to the buffer here?

steveiliop56: Whoops, fixed!

cmd/tinyauth/tinyauth.go
```diff
@@ -131,3 +154,68 @@ type themeBase struct{}
...
+
+type colors struct {
+	red    lipgloss.Style
+	green  lipgloss.Style
+	gray   lipgloss.Style
+	yellow lipgloss.Style
+	blue   lipgloss.Style
+}
+
+func getColors() colors {
```
Rycochet, line 167: This function is called several places to get the resulting struct - but the env vars will never change so it's doing extra work for it - could this have something like a global var above it, then have an early abort returning that if set, and then only need to set it once? (Or some better go version of that?)

steveiliop56: Not worth it. In go we cannot call functions if not in a function so, something like this wouldn't compile:

```go
var c getColors()
```

So the simple solution is to just call the function every time, not a big deal.

steveiliop56:
@coderabbitai review

Rycochet (dismissed):

Rycochet (approved):

PR #1042, opened 2026-07-25
fix: stop forwarding client Authorization header to downstream apps

evanwhitt:
setHeaders was piping the client's Authorization header straight through to the downstream app on every request. if someone auths via Basic Auth their creds end up in the backend. even with session-based auth, any Authorization header the client sent gets leaked.

the configured Basic Auth passthrough from ACLs is untouched, it still sets its own header separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented incoming authorization credentials from being exposed in proxy responses.
  * Responses now include only configured access-control headers and applicable authentication headers.


steveiliop56:
This is intended behavior. Tinyauth is not meant to "eat" any headers, only add new ones like the `Remote-` ones.

PR #1044, opened 2026-07-27
fix(acls): auth bypass via forward auth

scottmckendry:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of forwarded request paths for access control decisions.
  - Allow/block rules now correctly match intended prefixes and avoid substring/query-string mismatches.
  - Consistent root-path behavior for allow and block scenarios.
  - Invalid forwarded URIs and malformed path rules are safely rejected (deny by default).
- **Tests**
  - Expanded coverage for forwarded-path allow/block edge cases, including query-string handling and encoded/substring variations.

### DIFF
--- a/internal/controller/proxy_controller.go
+++ b/internal/controller/proxy_controller.go
@@ -9,14 +9,14 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/gin-gonic/gin"
+	"github.com/google/go-querystring/query"
+	"go.uber.org/dig"
+
 	"github.com/tinyauthapp/tinyauth/internal/model"
 	"github.com/tinyauthapp/tinyauth/internal/service"
 	"github.com/tinyauthapp/tinyauth/internal/utils"
 	"github.com/tinyauthapp/tinyauth/internal/utils/logger"
-	"go.uber.org/dig"
-
-	"github.com/gin-gonic/gin"
-	"github.com/google/go-querystring/query"
 )
 
 type AuthModuleType int
@@ -394,6 +394,10 @@ func (controller *ProxyController) getForwardAuthContext(c *gin.Context) (ProxyC
 
 	if err != nil {
 		return ProxyContext{}, fmt.Errorf("invalid x-forwarded-uri: %w", err)
+	}
+
+	if parsedURI.Path == "" {
+		parsedURI.Path = "/"
 	}
 
 	proto, ok := controller.getHeader(c, "x-forwarded-proto")
--- a/internal/controller/proxy_controller.go
+++ b/internal/controller/proxy_controller.go
@@ -390,6 +390,12 @@ func (controller *ProxyController) getForwardAuthContext(c *gin.Context) (ProxyC
 		return ProxyContext{}, errors.New("x-forwarded-uri not found")
 	}
 
+	parsedURI, err := url.ParseRequestURI(uri)
+
+	if err != nil {
+		return ProxyContext{}, fmt.Errorf("invalid x-forwarded-uri: %w", err)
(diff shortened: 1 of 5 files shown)


internal/service/access_controls_rules.go
```diff
@@ -180,33 +181,65 @@ type AuthEnabledRule struct {
...
+			continue
+		}
+
+		if strings.HasPrefix(path, configuredPath) {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
```
steveiliop56, line 212: @scottmckendry we are changing behavior here and possibly breaking things. The way ACLs for paths should work is the following:

## Normal whitelist

In case we have something like:

```
tinyauth.apps.foo.path.allow: /bar,/foo/bar,/hello
```

On every request we loop through the rules and check if the request URI has any prefix that matches. For example, `/bar` and all sub-paths such as `/bar/foo/bar` will be allowed. This is the current **intended** behavior.

## Regex whitelist

For regex whitelist there is no looping, the check for regex happens before the comma split and we just check if the rule starts and ends with slashes (`/`). For example this example is allowed:

```
tinyauth.apps.foo.path.allow: /\/foo|bar/
```

This is **not** allowed:

```
tinyauth.apps.foo.path.allow /\/foo|bar/,/\/test/
```

There is no need to allow such thing since your comma-seperated whitelist can be recreated using regex and users using regex will know what they are doing. Now, as for how regex is treated, it's a simple match (not a prefix match). It's the user's responsibility to know how to configure regex matching correctly, not ours.

Hopefully this clears things out a bit on how the parsing of the whitelist should work.

internal/controller/proxy_controller.go
```diff
@@ -390,6 +390,16 @@ func (controller *ProxyController) getForwardAuthContext(c *gin.Context) (ProxyC
...
 
+	parsedURI, err := url.ParseRequestURI(uri)
+
+	if err != nil {
+		return ProxyContext{}, fmt.Errorf("invalid x-forwarded-uri: %w", err)
+	}
+
+	if parsedURI.Path == "" {
+		parsedURI.Path = "/"
+	}
```
steveiliop56: We should do this for all auth module types and also strip out the query parameters immediately. Tinyauth won't need them in any part of the authentication chain anyway.

steveiliop56 (changes requested):

internal/service/access_controls_rules.go
scottmckendry: thanks, this is quite helpful actually. I think I've covered all scenarios now.

internal/controller/proxy_controller.go
scottmckendry: good call, done :)

internal/service/access_controls_rules.go
```diff
@@ -180,33 +181,56 @@ type AuthEnabledRule struct {
...
+		}
+
+		return regex.MatchString(path), nil
+	}
+
+	for _, configuredPath := range strings.Split(paths, ",") {
+		if strings.HasPrefix(path, strings.TrimSpace(configuredPath)) {
+			return true, nil
+		}
+	}
```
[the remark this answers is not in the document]
steveiliop56: @scottmckendry although this is 100% user error let's just apply the suggestion, doesn't cost much code-wiese.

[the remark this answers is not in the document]
scottmckendry: fixed in latest commit

steveiliop56:
Closing in favor of #1055, thanks @scottmckendry.

PR #1094, opened 2026-08-24
feat: support OIDC RP-initiated logout

norrs:
## Stack

- Current: #1094 — upstream provider RP-initiated logout
- Follow-up: #1102 — Tinyauth OIDC end-session endpoint (draft; merge after this PR)

OAuth-backed sessions currently only log out of Tinyauth. When the upstream provider keeps its SSO session, the next protected application access can immediately create a new Tinyauth session, so logout does not behave like an end-to-end sign-out for OIDC providers that support RP-initiated logout.

Add an optional OAuth provider logoutUrl for the OpenID Provider end_session_endpoint and keep the provider id_token server-side on the Tinyauth session. The logout handler now deletes the local session, builds the OP logout request with client_id, id_token_hint, post_logout_redirect_uri, and state, and returns that redirect to the frontend. The callback endpoint validates and restores the requested application return URL after the OP hop.

Persist oauth_id_token for SQLite, Postgres, memory, SQLC generated repositories, and store wrapper models so refreshed sessions retain the token. Add migrations for both database drivers. Update the logout page and quick actions menu to follow backend-provided redirect URLs while keeping Tinyauth redirect_uri separate from OIDC
post_logout_redirect_uri.

Cover the new behavior with controller tests for safe logout redirects, logout URL construction, and use of the server-side id_token. Enable TLS on the dev whoami route so the local Traefik setup exercises the secure-cookie and OIDC logout flow.

## Summary by CodeRabbit

* **New Features**
  * Added OpenID Connect single sign-out with configurable provider logout endpoints.
  * OAuth sign-out information is preserved across sessions and refreshes.
  * Added secure HTTPS access for the development diagnostic service.

* **Bug Fixes**
  * Improved protection against unsafe or cross-domain logout redirects, with a safe fallback destination.
  * Provider-supplied logout redirects are followed immediately when available.

* **Tests**
  * Added coverage for secure redirects and OpenID Connect logout handling.

### DIFF
--- a/frontend/src/components/quick-actions/quick-actions.tsx
+++ b/frontend/src/components/quick-actions/quick-actions.tsx
@@ -77,6 +77,10 @@ export const QuickActions = () => {
     }
     return "";
   })();
+  const logoutParams =
+    screenParams.redirect_uri && screenParams.login_for !== "oidc"
+      ? { login_for: "app", redirect_uri: screenParams.redirect_uri }
+      : undefined;
 
   const [isOpen, setIsOpen] = useState(false);
 
@@ -126,9 +130,7 @@ export const QuickActions = () => {
     // It is not the OIDC RP-Initiated Logout post_logout_redirect_uri.
     mutationFn: () =>
       axios.post("/api/user/logout", undefined, {
-        params: screenParams.redirect_uri
-          ? { redirect_uri: screenParams.redirect_uri }
-          : undefined,
+        params: logoutParams,
       }),
     mutationKey: ["logout"],
     onSuccess: (response) => {
--- a/frontend/src/components/quick-actions/quick-actions.tsx
+++ b/frontend/src/components/quick-actions/quick-actions.tsx
@@ -122,12 +122,25 @@ export const QuickActions = () => {
   })();
 
   const logoutMutation = useMutation({
-    mutationFn: () => axios.post("/api/user/logout"),
+    // redirect_uri is Tinyauth's existing application-navigation parameter.
+    // It is not the OIDC RP-Initiated Logout post_logout_redirect_uri.
+    mutationFn: () =>
+      axios.post("/api/user/logout", undefined, {
+        params: screenParams.redirect_uri
+          ? { redirect_uri: screenParams.redirect_uri }
+          : undefined,
+      }),
     mutationKey: ["logout"],
-    onSuccess: () => {
+    onSuccess: (response) => {
       toast.success(t("logoutSuccessTitle"), {
         description: t("logoutSuccessSubtitle"),
       });
+
+      const redirectUrl = response.data?.redirectUrl;
+      if (typeof redirectUrl === "string" && redirectUrl.length > 0) {
+        window.location.replace(redirectUrl);
+        return;
+      }
 
       redirectTimer.current = window.setTimeout(() => {
         window.location.replace(`/login${compiledParams}`);
--- a/frontend/src/components/quick-actions/quick-actions.tsx
+++ b/frontend/src/components/quick-actions/quick-actions.tsx
@@ -139,13 +139,12 @@ export const QuickActions = () => {
       });
 
       const redirectUrl = response.data?.redirectUrl;
-      if (typeof redirectUrl === "string" && redirectUrl.length > 0) {
-        window.location.replace(redirectUrl);
-        return;
-      }
-
       redirectTimer.current = window.setTimeout(() => {
-        window.location.replace(`/login${compiledParams}`);
+        if (typeof redirectUrl === "string" && redirectUrl.length > 0) {
+          window.location.replace(redirectUrl);
+        } else {
+          window.location.replace(`/login${compiledParams}`);
+        }
       }, 500);
     },
     onError: () => {
--- a/frontend/src/pages/logout-page.tsx
+++ b/frontend/src/pages/logout-page.tsx
@@ -36,15 +36,17 @@ export const LogoutPage = () => {
     }
     return "";
   })();
+  const logoutParams =
+    screenParams.redirect_uri && screenParams.login_for !== "oidc"
+      ? { login_for: "app", redirect_uri: screenParams.redirect_uri }
+      : undefined;
 
   const logoutMutation = useMutation({
     // redirect_uri is Tinyauth's existing application-navigation parameter.
     // It is not the OIDC RP-Initiated Logout post_logout_redirect_uri.
     mutationFn: () =>
       axios.post("/api/user/logout", undefined, {
-        params: screenParams.redirect_uri
-          ? { redirect_uri: screenParams.redirect_uri }
-          : undefined,
+        params: logoutParams,
       }),
     mutationKey: ["logout"],
     onSuccess: (response) => {
--- a/frontend/src/pages/logout-page.tsx
+++ b/frontend/src/pages/logout-page.tsx
@@ -55,13 +55,12 @@ export const LogoutPage = () => {
       });
 
       const redirectUrl = response.data?.redirectUrl;
-      if (typeof redirectUrl === "string" && redirectUrl.length > 0) {
-        window.location.replace(redirectUrl);
-        return;
-      }
-
       redirectTimer.current = window.setTimeout(() => {
-        window.location.replace(`/login${compiledParams}`);
+        if (typeof redirectUrl === "string" && redirectUrl.length > 0) {
+          window.location.replace(redirectUrl);
+        } else {
+          window.location.replace(`/login${compiledParams}`);
+        }
       }, 500);
     },
     onError: () => {
--- a/frontend/src/pages/logout-page.tsx
+++ b/frontend/src/pages/logout-page.tsx
@@ -38,12 +38,25 @@ export const LogoutPage = () => {
   })();
 
   const logoutMutation = useMutation({
-    mutationFn: () => axios.post("/api/user/logout"),
+    // redirect_uri is Tinyauth's existing application-navigation parameter.
+    // It is not the OIDC RP-Initiated Logout post_logout_redirect_uri.
+    mutationFn: () =>
+      axios.post("/api/user/logout", undefined, {
+        params: screenParams.redirect_uri
+          ? { redirect_uri: screenParams.redirect_uri }
+          : undefined,
+      }),
     mutationKey: ["logout"],
-    onSuccess: () => {
+    onSuccess: (response) => {
       toast.success(t("logoutSuccessTitle"), {
         description: t("logoutSuccessSubtitle"),
       });
+
+      const redirectUrl = response.data?.redirectUrl;
+      if (typeof redirectUrl === "string" && redirectUrl.length > 0) {
+        window.location.replace(redirectUrl);
+        return;
+      }
 
       redirectTimer.current = window.setTimeout(() => {
         window.location.replace(`/login${compiledParams}`);
--- a/internal/controller/user_controller.go
+++ b/internal/controller/user_controller.go
@@ -13,6 +13,7 @@ import (
 	"github.com/tinyauthapp/tinyauth/internal/service"
 	"github.com/tinyauthapp/tinyauth/internal/utils"
 	"github.com/tinyauthapp/tinyauth/internal/utils/logger"
+	"github.com/tinyauthapp/tinyauth/pkg/validators"
 	"go.uber.org/dig"
 
 	"github.com/gin-gonic/gin"
@@ -30,6 +31,7 @@ type TotpRequest struct {
 
 type UserController struct {
 	log     *logger.Logger
+	config  *model.Config
 	runtime *model.RuntimeConfig
 	auth    *service.AuthService
 }
@@ -38,6 +40,7 @@ type UserControllerInput struct {
 	dig.In
 
 	Log           *logger.Logger
+	StaticConfig  *model.Config
 	RuntimeConfig *model.RuntimeConfig
 	RouterGroup   *gin.RouterGroup `name:"apiRouterGroup"`
 	AuthService   *service.AuthService
@@ -46,6 +49,7 @@ type UserControllerInput struct {
 func NewUserController(i UserControllerInput) *UserController {
 	controller := &UserController{
 		log:     i.Log,
+		config:  i.StaticConfig,
 		runtime: i.RuntimeConfig,
 		auth:    i.AuthService,
 	}
@@ -321,31 +325,41 @@ func (controller *UserController) safeLogoutRedirect(raw string) string {
 		return fallback
 	}
 
-	target, err := url.Parse(raw)
-	if err != nil || target.Host == "" || target.User != nil {
-		return fallback
-	}
-	if target.Scheme != "http" && target.Scheme != "https" {
-		return fallback
-	}
-
 	appURL, err := url.Parse(controller.runtime.AppURL)
 	if err != nil {
 		return fallback
 	}
-	if appURL.Scheme == "https" && target.Scheme != "https" {
+
+	allowedSchemes := []string{"http", "https"}
+	if appURL.Scheme == "https" {
+		allowedSchemes = []string{"https"}
+	}
+
+	schemeValidator := validators.NewDomainValidator(validators.DomainValidatorOptions{
+		WithScheme:     true,
+		AllowedSchemes: allowedSchemes,
+	})
+	hostname, err := schemeValidator.SafeHostname(raw)
+	if err != nil {
 		return fallback
 	}
 
-	targetHost := strings.ToLower(target.Hostname())
-	appHost := strings.ToLower(appURL.Hostname())
-	if targetHost == appHost {
+	domainValidator := validators.NewDomainValidator(validators.DomainValidatorOptions{
+		WithPort: true,
+	})
+	err = domainValidator.Validate(raw, controller.runtime.AppURL)
+	if err == nil {
 		return raw
 	}
 
-	cookieDomain := strings.TrimPrefix(strings.ToLower(controller.runtime.CookieDomain), ".")
-	if cookieDomain != "" &&
-		(targetHost == cookieDomain || strings.HasSuffix(targetHost, "."+cookieDomain)) {
+	if !errors.Is(err, validators.ErrHostnameMismatch) ||
+		controller.config == nil ||
+		!controller.config.Auth.SubdomainsEnabled {
+		return fallback
+	}
+
+	cookieDomain := strings.ToLower(controller.runtime.CookieDomain)
+	if hostname == cookieDomain || strings.HasSuffix(hostname, "."+cookieDomain) {
 		return raw
 	}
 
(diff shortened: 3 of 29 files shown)


internal/controller/user_controller.go
```diff
@@ -227,51 +230,177 @@ func (controller *UserController) loginHandler(c *gin.Context) {
...
+		return "", fmt.Errorf("unsupported logout URL scheme")
+	}
+
+	query := logoutURL.Query()
+	if provider.ClientID != "" {
+		query.Set("client_id", provider.ClientID)
+	}
+	if idToken != "" {
+		query.Set("id_token_hint", idToken)
+	}
```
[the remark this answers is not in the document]
norrs: Pushed in https://github.com/tinyauthapp/tinyauth/compare/d56f4d1bd636623be102d05fd913c6cad8982884..dfcf8110acde523ef254fb1448f4b366663ab305

internal/controller/user_controller.go
steveiliop56, line 1: Throughout the entire file I notice that you use a separate variable for each error. There is no need for such thing. You can just do:

```go
err := doSomeAction()

if err != nil {
  return err
}

err = doSomeOtherAction()
...
```

internal/controller/user_controller.go
```diff
@@ -227,51 +230,183 @@ func (controller *UserController) loginHandler(c *gin.Context) {
...
 	}
 
-	cookie, err := controller.auth.DeleteSession(c, uuid)
+	// If middleware context is missing, fall back to the just-loaded session
+	// provider. If there is no session metadata either, a deployment with exactly
+	// one OAuth provider can still terminate that provider's SSO session.
+	if providerID == "" && contextErr != nil && isSessionOAuthProvider(sessionProviderID) {
+		providerID = sessionProviderID
+	}
+	if providerID == "" && contextErr != nil && sessionProviderID == "" && len(controller.runtime.OAuthProviders) == 1 {
```
steveiliop56: I would check if the context is `nil` here instead of checking the errors.

internal/controller/user_controller.go
```diff
@@ -227,51 +230,183 @@ func (controller *UserController) loginHandler(c *gin.Context) {
...
+func isSessionOAuthProvider(providerID string) bool {
+	switch providerID {
+	case "", "local", "ldap", "tailscale":
+		return false
+	default:
+		return true
+	}
+}
 
+func (controller *UserController) safeLogoutRedirect(raw string) string {
```
steveiliop56, line 329: Please use the domain validator for any validating logic. See `isRedirectSafe` in the OAuth controller.

internal/controller/user_controller.go
```diff
@@ -227,51 +230,183 @@ func (controller *UserController) loginHandler(c *gin.Context) {
...
+	sessionProviderID := ""
 	uuid, err := c.Cookie(controller.runtime.SessionCookieName)
+	if err == nil {
+		session, sessionErr := controller.auth.GetSession(c, uuid)
+		if sessionErr != nil {
+			controller.log.App.Warn().Err(sessionErr).Msg("Failed to get session during logout, continuing without session-backed logout metadata")
+		} else {
+			idToken = session.OAuthIDToken
+			sessionProviderID = session.Provider
+		}
```
steveiliop56: There is no reason to do a duplicate database lookup for the OAuth ID token. The context middleware can do that. Please move the ID token into a field in the OAuth-specific context (like the sub).

internal/controller/user_controller.go
```diff
@@ -227,51 +230,183 @@ func (controller *UserController) loginHandler(c *gin.Context) {
...
 		return
 	}
 
-	cookie, err := controller.auth.DeleteSession(c, uuid)
+	// If middleware context is missing, fall back to the just-loaded session
+	// provider. If there is no session metadata either, a deployment with exactly
+	// one OAuth provider can still terminate that provider's SSO session.
+	if providerID == "" && contextErr != nil && isSessionOAuthProvider(sessionProviderID) {
+		providerID = sessionProviderID
+	}
```
steveiliop56: No need to do this. The context middleware will always return the same session as the lookup. After https://github.com/tinyauthapp/tinyauth/pull/1094/changes#r3864349886, the session lookup won't even be needed.

internal/controller/user_controller.go
```diff
@@ -227,51 +230,183 @@ func (controller *UserController) loginHandler(c *gin.Context) {
...
+	if providerID == "" && contextErr != nil && sessionProviderID == "" && len(controller.runtime.OAuthProviders) == 1 {
+		for id := range controller.runtime.OAuthProviders {
+			providerID = id
+		}
+	}
```
steveiliop56: I am not a fan of this. It seems non-deterministic behavior. I would recommend just skipping the OAuth logout if we can't determine what OAuth provider triggered the logout.

internal/controller/user_controller.go
```diff
@@ -227,51 +230,183 @@ func (controller *UserController) loginHandler(c *gin.Context) {
...
+	provider, ok := controller.runtime.OAuthProviders[providerID]
+	if ok && provider.LogoutURL != "" {
+		// OpenID Connect RP-Initiated Logout 1.0:
+		// https://openid.net/specs/openid-connect-rpinitiated-1_0-final.html#RPLogout
+		//
+		// OP-facing standardized parameters:
+		//   id_token_hint
+		//   post_logout_redirect_uri
+		//   state
+		callbackURL := strings.TrimRight(controller.runtime.AppURL, "/") + "/api/user/logout/callback"
```
steveiliop56: No need to modify the runtime app URL in any way, it's already in the form of `scheme://host:port`, it does not include a path.

internal/controller/user_controller.go
```diff
@@ -227,51 +230,183 @@ func (controller *UserController) loginHandler(c *gin.Context) {
...
+	logoutURL, err := url.Parse(provider.LogoutURL)
+	if err != nil || logoutURL.Host == "" {
+		return "", fmt.Errorf("invalid logout URL")
+	}
+	if logoutURL.Scheme != "http" && logoutURL.Scheme != "https" {
+		return "", fmt.Errorf("unsupported logout URL scheme")
+	}
+	if logoutURL.Scheme == "http" && !provider.Insecure {
+		return "", fmt.Errorf("insecure logout URL requires insecure OAuth provider")
+	}
```
steveiliop56: This is not the case. Insecure just means trust the self-signed certificate, not run in HTTP. This check can be removed.

frontend/src/components/quick-actions/quick-actions.tsx
```diff
@@ -122,13 +122,26 @@ export const QuickActions = () => {
   })();
 
   const logoutMutation = useMutation({
-    mutationFn: () => axios.post("/api/user/logout"),
+    // redirect_uri is Tinyauth's existing application-navigation parameter.
+    // It is not the OIDC RP-Initiated Logout post_logout_redirect_uri.
+    mutationFn: () =>
+      axios.post("/api/user/logout", undefined, {
+        params: screenParams.redirect_uri
+          ? { redirect_uri: screenParams.redirect_uri }
```
steveiliop56: Here, we need to check if the parameters are from an OIDC request (`login_for` will be `oidc`) and if they are not, we need to also specify `login_for=app` to Tinyauth so it knows where it's redirecting to after the logout.

frontend/src/pages/logout-page.tsx
```diff
@@ -38,13 +38,26 @@ export const LogoutPage = () => {
...
 
   const logoutMutation = useMutation({
-    mutationFn: () => axios.post("/api/user/logout"),
+    // redirect_uri is Tinyauth's existing application-navigation parameter.
+    // It is not the OIDC RP-Initiated Logout post_logout_redirect_uri.
+    mutationFn: () =>
+      axios.post("/api/user/logout", undefined, {
+        params: screenParams.redirect_uri
+          ? { redirect_uri: screenParams.redirect_uri }
+          : undefined,
```
steveiliop56: See https://github.com/tinyauthapp/tinyauth/pull/1094/changes#r3864396759

internal/controller/user_controller_sso_logout_test.go
```diff
@@ -0,0 +1,278 @@
...
+	assert.Equal(
+		t,
+		"https://auth.example.com",
+		controller.safeLogoutRedirect("javascript:alert(1)"),
+	)
+	assert.Equal(
+		t,
+		"https://auth.example.com",
+		controller.safeLogoutRedirect("http://app.example.com/"),
+	)
```
steveiliop56: No need to re-test the redirect checking logic since it will be handled by the validator which already includes a wide variety of tests.

internal/controller/user_controller_sso_logout_test.go
```diff
@@ -0,0 +1,278 @@
...
+	assert.Equal(t, "http", parsed.Scheme)
+	assert.Equal(t, "id.example.com", parsed.Host)
+	assert.Equal(t, "id-token", parsed.Query().Get("id_token_hint"))
+	assert.Equal(t, "client-id", parsed.Query().Get("client_id"))
+	assert.Equal(
+		t,
+		"https://auth.example.com/api/user/logout/callback",
+		parsed.Query().Get("post_logout_redirect_uri"),
+	)
+}
```
steveiliop56: Please use a table-driven testing approach the rest of the controllers and services use.

steveiliop56 (changes requested):

internal/controller/user_controller.go
norrs: I've aligned with the code style to re-use `err` between blocks.

frontend/src/components/quick-actions/quick-actions.tsx
```diff
@@ -122,13 +126,24 @@ export const QuickActions = () => {
...
+    onSuccess: (response) => {
       toast.success(t("logoutSuccessTitle"), {
         description: t("logoutSuccessSubtitle"),
       });
 
+      const redirectUrl = response.data?.redirectUrl;
+      if (typeof redirectUrl === "string" && redirectUrl.length > 0) {
+        window.location.replace(redirectUrl);
+        return;
+      }
```
norrs: Should we consider moving this into the function below with setTimeout()?

frontend/src/pages/logout-page.tsx
```diff
@@ -36,15 +36,30 @@ export const LogoutPage = () => {
...
       toast.success(t("logoutSuccessTitle"), {
         description: t("logoutSuccessSubtitle"),
       });
 
+      const redirectUrl = response.data?.redirectUrl;
+      if (typeof redirectUrl === "string" && redirectUrl.length > 0) {
+        window.location.replace(redirectUrl);
+        return;
+      }
+
```
norrs: Ditto, consider moving this into window.setTimeout?

internal/controller/user_controller.go
norrs: Addressed in [22a0e8eb66fa](https://github.com/tinyauthapp/tinyauth/commit/22a0e8eb66fa60fef0f33d57f9e457daa82e4fb5), [860d8895057a](https://github.com/tinyauthapp/tinyauth/commit/860d8895057a5d56cdd782bcc91ac92f3ae2a468), and [24f3d824703c](https://github.com/tinyauthapp/tinyauth/commit/24f3d824703c3bc892ba7b4a797dc99bc7fe62a5): the logout flow now uses `err` instead of separate `sessionErr`, `deleteErr`, `contextErr`, and `buildErr` variables. The final commit is limited to the `buildErr` → `err` rename.

internal/controller/user_controller.go
norrs: Addressed in [22a0e8eb66fa](https://github.com/tinyauthapp/tinyauth/commit/22a0e8eb66fa60fef0f33d57f9e457daa82e4fb5) and [860d8895057a](https://github.com/tinyauthapp/tinyauth/commit/860d8895057a5d56cdd782bcc91ac92f3ae2a468). The logout flow checks `userContext != nil`; if context retrieval returns an error, `userContext` is explicitly set to `nil` before it is used.

internal/controller/user_controller.go
norrs: Addressed in [61efe839398f](https://github.com/tinyauthapp/tinyauth/commit/61efe839398fff24b7d821cc8c4beaa038d0c443). `safeLogoutRedirect` now uses the shared domain validator for scheme, hostname, and port validation, following the OAuth controller’s approach and respecting the subdomain setting.

internal/controller/user_controller.go
norrs: Addressed in [22a0e8eb66fa](https://github.com/tinyauthapp/tinyauth/commit/22a0e8eb66fa60fef0f33d57f9e457daa82e4fb5). Added `IDToken` to `OAuthContext`, populated it from the session when constructing the context, and made logout read it from that context. The duplicate session lookup is removed.

internal/controller/user_controller.go
norrs: Addressed in [22a0e8eb66fa](https://github.com/tinyauthapp/tinyauth/commit/22a0e8eb66fa60fef0f33d57f9e457daa82e4fb5). Removed the extra session lookup and session-backed provider fallback; logout gets the provider and ID token from the middleware-populated OAuth context.

internal/controller/user_controller.go
norrs: Addressed in [22a0e8eb66fa](https://github.com/tinyauthapp/tinyauth/commit/22a0e8eb66fa60fef0f33d57f9e457daa82e4fb5). Removed the single-provider guess. If the OAuth provider cannot be determined from the user context, provider logout is skipped while local logout still proceeds.

internal/controller/user_controller.go
norrs: Addressed in [424449cb6945](https://github.com/tinyauthapp/tinyauth/commit/424449cb69458383feacdba1ba43038ab438ccd1). The callback is now built directly as `controller.runtime.AppURL + "/api/user/logout/callback"`, without normalizing the runtime URL.

internal/controller/user_controller.go
norrs: Addressed in [55ac9d4a1b72](https://github.com/tinyauthapp/tinyauth/commit/55ac9d4a1b724d28fb390e7130a034ed830216d2). Provider logout URLs must now use HTTPS regardless of `provider.Insecure`; the HTTP exception has been removed.

frontend/src/components/quick-actions/quick-actions.tsx
norrs: Addressed in [0cf0dee8709c](https://github.com/tinyauthapp/tinyauth/commit/0cf0dee8709c985c90d8e6e042f88d89fe6af7dc) and [e5e5b27499e0](https://github.com/tinyauthapp/tinyauth/commit/e5e5b27499e0afa1414df17222fd9345e19c64a6). Quick actions omit app redirect parameters for `login_for=oidc`; otherwise an app redirect is sent with `login_for=app`. The backend also only accepts the app redirect parameter when `login_for=app` is explicit.

frontend/src/pages/logout-page.tsx
norrs: Addressed in [0cf0dee8709c](https://github.com/tinyauthapp/tinyauth/commit/0cf0dee8709c985c90d8e6e042f88d89fe6af7dc) and [e5e5b27499e0](https://github.com/tinyauthapp/tinyauth/commit/e5e5b27499e0afa1414df17222fd9345e19c64a6). The logout page now applies the same check as quick actions: OIDC parameters are not forwarded as app logout redirects, and app redirects include `login_for=app`. The backend enforces that intent as well.

internal/controller/user_controller_sso_logout_test.go
norrs: Addressed in [61efe839398f](https://github.com/tinyauthapp/tinyauth/commit/61efe839398fff24b7d821cc8c4beaa038d0c443). Removed the controller tests that duplicated the shared domain validator’s redirect-validation coverage.

internal/controller/user_controller_sso_logout_test.go
norrs: Addressed in [084520808020](https://github.com/tinyauthapp/tinyauth/commit/08452080802051a509f39aee5f102594ad3b3743). The SSO logout and logout URL builder tests now use table-driven cases. Follow-up [32c415257ec9](https://github.com/tinyauthapp/tinyauth/commit/32c415257ec90e506b65e44b91c592b6e824ee63) uses `test` for the loop variable to match the project’s existing naming.

frontend/src/components/quick-actions/quick-actions.tsx
norrs: I moved it: https://github.com/tinyauthapp/tinyauth/pull/1094/commits/1efe7704204c812f6d6a4cd610d208c11be1ed75

frontend/src/pages/logout-page.tsx
norrs: I moved it: https://github.com/tinyauthapp/tinyauth/pull/1094/commits/1efe7704204c812f6d6a4cd610d208c11be1ed75

pkg/validators/domain_validator.go
steveiliop56, line 1: Any reason for failing on userinfo? We can just ignore it.

internal/controller/user_controller.go
```diff
@@ -227,51 +234,168 @@ func (controller *UserController) loginHandler(c *gin.Context) {
...
+	} else {
 		controller.log.App.Error().Err(err).Msg("Error retrieving session cookie on logout")
-		c.JSON(500, gin.H{
-			"status":  500,
+		c.JSON(http.StatusInternalServerError, gin.H{
+			"status":  http.StatusInternalServerError,
 			"message": "Internal Server Error",
 		})
 		return
 	}
```
steveiliop56, line 287: I would avoid the `if else if else` statements. The logic I'd do is the following (used in other controllers as well):

```
cookie, err = get cookie

if err != nil
    if error is no cookie
        log
    else
        fail
else
    ...
```

internal/controller/user_controller.go
```diff
@@ -227,51 +234,168 @@ func (controller *UserController) loginHandler(c *gin.Context) {
...
+		query.Set("client_id", provider.ClientID)
+	}
+	if idToken != "" {
+		query.Set("id_token_hint", idToken)
+	}
+	query.Set("post_logout_redirect_uri", callbackURL)
+	if state != "" {
+		query.Set("state", state)
+	}
+	logoutURL.RawQuery = query.Encode()
```
steveiliop56, line 396: Maybe we should use the Google query library instead and use a struct?

internal/controller/user_controller.go
```diff
@@ -227,51 +234,168 @@ func (controller *UserController) loginHandler(c *gin.Context) {
...
-		})
-		return
+		return fallback
 	}
 
-	context, err := new(model.UserContext).NewFromGin(c)
+	allowedSchemes := []string{"http", "https"}
+	if appURL.Scheme == "https" {
+		allowedSchemes = []string{"https"}
+	}
```
steveiliop56, line 343: Better to allow both regardless of the app URL status.

steveiliop56 (changes requested):

PR #1101, opened 2026-08-28
feat(ui): allow configuring background via css instead of an image

flbraun:
Thank you for Tinyauth! This is my first contribution to this project.

The default background image (foggy forest) transfers ~550kB over the wire on a cold cache. Many of my users run on spotty and slow connections and report that loading the background image causes severe UI flickering.

I'm aware that one may configure a custom, smaller background image via the `TINYAUTH_UI_BACKGROUNDIMAGE` setting, but I think I've found an even better solution for this scenario: The new `TINYAUTH_UI_BACKGROUNDCSS` config allows specifying the value of the backdrop's `background` CSS property. With modern browser's CSS capabilities, this allows for a surprising amount of creativity and doesn't require **any** additional network transfers.

For example, setting `TINYAUTH_UI_BACKGROUNDCSS="linear-gradient(135deg, #03045e 0%, #0077b6 50%, #00b4d8 100%)"` makes the UI look like this:
after

Your thoughts about this? Is anything missing from this PR?

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the UI background with a custom CSS value.
  * Custom background CSS takes precedence over the configured background image.
  * Added the `TINYAUTH_UI_BACKGROUNDCSS` environment variable for background customization.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -205,6 +205,8 @@ TINYAUTH_UI_TITLE="Tinyauth"
 TINYAUTH_UI_FORGOTPASSWORDMESSAGE="You can change your password by changing the configuration."
 # Path to the background image.
 TINYAUTH_UI_BACKGROUNDIMAGE="/background.jpg"
+# Value of the CSS background property, takes precedence over the background image.
+TINYAUTH_UI_BACKGROUNDCSS=
 # Enable UI warnings.
 TINYAUTH_UI_WARNINGSENABLED=true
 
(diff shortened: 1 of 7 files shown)


steveiliop56:
Haha sounds like a fun feature. Code looks good to me, I will handle the conflicts. Thanks!

steveiliop56 (approved):

Rycochet (approved):

PR #1102, opened 2026-08-30
feat: expose OIDC end-session endpoint

norrs:
## Stack

- Parent: #1094
- Current: expose the OIDC end-session endpoint

This PR depends on #1094 and must merge after it. Because the branches are on a fork, GitHub requires this PR to target `main`; its Files view will include the parent until #1094 merges. Review the endpoint-only change here:

https://github.com/norrs/tinyauth/compare/feat/sso-logout...feat/oidc-end-session

## Summary

- advertise `end_session_endpoint` in OIDC discovery
- validate downstream ID token hints and registered post-logout redirect URIs
- reuse the upstream provider logout cascade for OIDC clients
- protect provider callbacks with short-lived, one-time tickets
- cover the endpoint, validation, redirects, discovery metadata, and logout cascade

## Verification

- `make test`
- `make vet`
- `go test -count=1 ./internal/controller ./internal/service`


steveiliop56:
*Note to myself*

Will need to review the https://openid.net/specs/openid-connect-rpinitiated-1_0.html document before reviewing this.

Issue #496, opened 2025-12-04
[BUG] Tinyauth doesn't retrieve groups with AzureAD


naub1n:
**Describe the bug**
Hi,

Tinyauth doesn't retrieve groups with AzureAD.
I success to log in but I can't access to app because tinyauth don't find groups with my account.

<img width="386" height="213" alt="Image" src="https://github.com/user-attachments/assets/a429147c-7208-40f3-b0bc-161a32c5cbda" />

TinyAuth logs :

```
2025-12-04T09:39:14Z DBG internal/controller/oauth_controller.go:188 > No preferred username from OAuth provider, using pseudo username

Email=XXXXXXXX Name=AUBIN Nicolas OAuthGroups= OAuthName=AzureAd Provider=generic TotpPending=false UUID= Username=XXXXXXXX

2025-12-04T09:39:14Z TRC internal/controller/oauth_controller.go:201 > Creating session cookie session_cookie=
```
**`OAuthGroups` is empty !**

My container config:

```
PROVIDERS_GENERIC_CLIENT_ID: <client_id>
PROVIDERS_GENERIC_CLIENT_SECRET: <client_secret>
PROVIDERS_GENERIC_AUTH_URL: https://login.microsoftonline.com/<tenant_id>/oauth2/v2.0/authorize
PROVIDERS_GENERIC_TOKEN_URL: https://login.microsoftonline.com/<tenant_id>/oauth2/v2.0/token
PROVIDERS_GENERIC_USER_INFO_URL: https://graph.microsoft.com/oidc/userinfo
PROVIDERS_GENERIC_REDIRECT_URL: https://<my_app_host>/api/oauth/callback/generic
PROVIDERS_GENERIC_SCOPES: "openid profile email"
PROVIDERS_GENERIC_NAME: AzureAd
```

My id_token gets groups:

<img width="676" height="593" alt="Image" src="https://github.com/user-attachments/assets/407870ef-cf6c-4015-b921-484ef4c4f4f2" />

**Device (please complete the following information):**
 - OS: Debian
 - Browser Firefox
 - Tinyauth V4.1.0
 - Docker 27.4.1 (rootless)

naub1n:
TinyAuth seems to read groups with `USER_INFO_URL`, AzureAd doesn't send groups in `https://graph.microsoft.com/oidc/userinfo`. 😢 . It is right ?

steveiliop56:
[...]

PR #1126, opened 2026-09-11
feat: support X-Api-Key header for basic authentication

eGamesAPI:
# PR: feat: support X-Api-Key header for basic authentication

## Target
`tinyauthapp/tinyauth` ← branch `feat/x-api-key-header`

## Title
feat: support X-Api-Key header for basic authentication

### Problem

When an application behind TinyAuth authenticates its own clients through the
`Authorization` header (e.g. `Authorization: Bearer <token>` — API panels such
as Remnawave, Grafana-style dashboards, etc.), a client cannot send both
TinyAuth basic credentials and the application token: the standard basic auth
scheme and the application's bearer token compete for the same single header.

The result: such apps either have to leave their API routes completely
unprotected at the proxy level, or clients lose token authentication.

### Solution

Accept TinyAuth basic credentials in a dedicated `X-Api-Key` header:

```
X-Api-Key: Basic base64(username:password)
Authorization: Bearer <application token>   ← passes through untouched
```

Behavior (mirrors the semantics already shipped and battle-tested in the
Remnawave community fork maposia/tinyauth):

1. If `X-Api-Key` is present — it is the only source of basic credentials:
   - valid `Basic base64(user:pass)` → authenticated;
   - malformed value or a different scheme → **rejected, no fallback**
     (a half-configured client must fail loudly, not silently degrade).
2. If `X-Api-Key` is absent — standard `Authorization` basic auth, exactly
   as before (zero behavior change for existing deployments).
3. The original `Authorization` header is never consumed or modified when
   `X-Api-Key` is used, so the downstream application receives its bearer
   token intact.

### Use cases

- Remnawave panel behind TinyAuth: the panel API uses Bearer tokens; users
  authenticate to TinyAuth with X-Api-Key in the same request
  (documented at https://docs.rw/security/tinyauth-for-nginx).
- Any bearer-token API behind a TinyAuth-protected route.

### Changes

- `internal/middleware/context_middleware.go`: try `X-Api-Key` before the
  standard `BasicAuth()`; on a malformed X-Api-Key reject without fallback.
- `internal/middleware/context_middleware_test.go`: table tests for the
  four cases (valid key / missing / malformed / wrong scheme).

### Not included (deliberately)

- No config flag: the fallback keeps existing deployments byte-compatible;
  a flag would only add a way to break the documented combination.
- No support for non-Basic schemes inside X-Api-Key (bearer keys etc.) —
  out of scope, keeps the surface minimal.

### Prior art

The same feature has been running in production via the
`ghcr.io/maposia/remnawave-tinyauth` fork (commit 2e94981, Dec 2025) and is
referenced by the official Remnawave guide for nginx integration
([remnawave/panel PR #496](https://github.com/remnawave/panel/pull/496)). Upstreaming it removes the need for the fork.

## Bug Fixes
- Added support for authentication through the `X-Tinyauth-Authorization` header.
- This header now takes precedence over the standard `Authorization` header.
- Malformed, unsupported, or empty values return `401 Unauthorized` without falling back to `Authorization`.
- Valid Basic authentication credentials are processed correctly.

## Tests
- Expanded coverage for valid authentication, header precedence, invalid formats, empty values, and fallback prevention.

### DIFF
--- a/internal/middleware/context_middleware.go
+++ b/internal/middleware/context_middleware.go
@@ -2,6 +2,7 @@ package middleware
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"net/http"
@@ -94,23 +95,28 @@ func (m *ContextMiddleware) Middleware() gin.HandlerFunc {
 			}
 		}
 
-		username, password, ok := c.Request.BasicAuth()
-
-		if ok {
-			userContext, headers, err := m.basicAuth(username, password)
-
-			if err != nil {
-				m.log.App.Error().Msgf("Error authenticating basic auth: %v", err)
-				c.Next()
+		// X-Api-Key takes priority when present: it lets a client carry
+		// TinyAuth basic credentials alongside an application token in the
+		// Authorization header (e.g. "Authorization: Bearer ..." APIs behind
+		// the proxy). A malformed or non-Basic X-Api-Key is rejected WITHOUT
+		// falling back to Authorization — a half-configured client must fail
+		// loudly instead of silently degrading.
+		if apiKey := c.Request.Header.Get("X-Api-Key"); apiKey != "" {
+			username, password, ok := parseAPIKeyBasicAuth(apiKey)
+			if !ok {
+				m.log.App.Debug().Msg("Invalid basic auth in X-Api-Key header")
+				c.AbortWithStatus(http.StatusUnauthorized)
 				return
 			}
 
-			for k, v := range headers {
-				c.Header(k, v)
-			}
+			m.handleBasicAuth(c, username, password)
+			return
+		}
 
-			c.Set("context", userContext)
-			c.Next()
+		username, password, ok := c.Request.BasicAuth()
+
+		if ok {
+			m.handleBasicAuth(c, username, password)
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


internal/middleware/context_middleware.go
```diff
@@ -94,6 +95,38 @@ func (m *ContextMiddleware) Middleware() gin.HandlerFunc {
...
 
+		// X-Api-Key takes priority when present: it lets a client carry
+		// TinyAuth basic credentials alongside an application token in the
+		// Authorization header (e.g. "Authorization: Bearer ..." APIs behind
+		// the proxy). A malformed or non-Basic X-Api-Key is rejected WITHOUT
+		// falling back to Authorization: a half-configured client must fail
+		// loudly instead of silently degrading. Presence is checked via the
+		// header map, because Get cannot tell an absent header from an
+		// explicitly empty one.
+		if apiKeyHeaders := c.Request.Header["X-Api-Key"]; len(apiKeyHeaders) > 0 {
```
steveiliop56: I would prefer a Tinyauth-owned header like `X-Tinyauth-Authorization`. We are not using API keys so `X-Api-Key` sounds misleading.

steveiliop56 (changes requested):
Looks good, just a small note on the header name. Also I would remove the excessive comments. I believe the code is pretty self-explanatory.

eGamesAPI:
Done: renamed the header to X-Tinyauth-Authorization everywhere (middleware, tests) and dropped the verbose comments. Thanks for the review!

manysuq (commented):
Thanks for the rename and for dropping the verbose comments — the updated version reads well. I re-reviewed the current head and ran it locally:

- `go test ./internal/middleware/...` passes on the PR head
- `go vet` is clean
- The error path (`m.basicAuth` error → `c.Next()` unauthenticated) correctly mirrors the existing `Authorization` flow, and the malformed / non-Basic / explicitly-empty cases rejecting **without** fallback to `Authorization` are exactly the right tests to have.

A few small things, none blocking:

1. **Naming follow-up on the rename:** `parseAPIKeyBasicAuth` still says "APIKey" while the header is now `X-Tinyauth-Authorization`. Something like `parseBasicAuthHeaderValue` would keep the two in sync.

2. **Test suggestion:** a case pinning a password that contains a colon (e.g. base64 of `user:pa:ss`) — `strings.Cut` splits at the first colon so the password keeps everything after it, matching `http.Request.BasicAuth()`, but it would be nice to lock that in since it's easy to break silently in a refactor.

3. **Docs:** the [docs repo](https://github.com/tinyauthapp/docs) has a section on authenticating to apps with basic auth (this PR's exact use case). Since the header is user-facing, should the docs get a matching paragraph? Happy to open that PR once the header name is final.

Note: this review was prepared with AI-assisted code analysis, per [AI_POLICY.md](https://github.com/tinyauthapp/tinyauth/blob/main/AI_POLICY.md) — all findings were verified by me: tests and vet run locally against the PR head, surrounding code read on `main`.

eGamesAPI:
Thanks for the thorough re-review! Points 1 and 2 are addressed in b328da0:

1. Renamed `parseAPIKeyBasicAuth` to `parseBasicAuthHeaderValue` so the helper stays in sync with the header name.
2. Added a test with a password containing a colon (`user:pa:ss` style, a local `colonuser` seeded in the middleware test) asserting the split keeps everything after the first colon, matching `http.Request.BasicAuth()`.

On (3): `X-Tinyauth-Authorization` is final, the rename was requested by the maintainer, so yes, please feel free to open the docs PR.

manysuq:
Docs PR is up: tinyauthapp/docs#135 — documents X-Tinyauth-Authorization in the Basic Authentication reference (precedence + 401 on malformed). Verified the wording against the b328da0 head; happy to adjust once maintainer gives the final call on the header name.

Issue #1130, opened 2026-09-13
[BUG] Creating an OIDC client with a hyphen in its name generates invalid systemd environment variables


LukasSkywalker:
### Describe the Bug

Creating a new OIDC client via the command line tries to turn the client name into an environment variable. If the name contains a hyphen, this name is turned into a variable name with a hyphen similar to this: `TINYAUTH_OIDC_CLIENTS_MY-APP_CLIENTID`.
When the variable is placed in a file and loaded via `systemd`s `EnvironmentFile=`, lines with variable names containing hyphens are ignored: https://man7.org/linux/man-pages/man5/systemd.exec.5.html#ENVIRONMENT
Disallowing hyphens would fix that issue.

### How to Reproduce

1. Run `./tinyauth oidc create my-app`
2. It displays the environment variable `TINYAUTH_OIDC_CLIENTS_MY-APP_CLIENTID`
3. When running `tinyauth` via `systemd`, this variable is ignored and OIDC does not work.

### Expected Behavior

It should display an error, because an environment variable with a dash is invalid in systemd.

### Additional Context

The variables are ignored when they contain a hyphen:

```bash
root@tinyauth:~# cat /opt/tinyauth/.env
TINYAUTH_DATABASE_PATH=/opt/tinyauth/database.db
TINYAUTH_AUTH_USERS='foo@example.com:$$2a$$10$$redacted123'
TINYAUTH_APPURL=https://auth.example.com
TINYAUTH_OIDC_CLIENTS_MY-APP_CLIENTID=foo
TINYAUTH_OIDC_CLIENTS_MY-APP_CLIENTSECRET=bar
TINYAUTH_OIDC_CLIENTS_MY-APP_NAME=baz

root@tinyauth:~# systemd-run --wait --pipe -p EnvironmentFile=/opt/tinyauth/.env /usr/bin/env | grep -o '^TINYAUTH_[^=]*'
Running as unit: run-p847-i2375553.service; invocation ID: 8bb8043490f3415a8ba4744f5cad4918
TINYAUTH_DATABASE_PATH
TINYAUTH_AUTH_USERS
TINYAUTH_APPURL
Finished with result: success
```

But are kept when they don't:

```bash
root@tinyauth:~# cat /opt/tinyauth/.env 
TINYAUTH_DATABASE_PATH=/opt/tinyauth/database.db
TINYAUTH_AUTH_USERS='foo@example.com:$$2a$$10$$redacted123'
[...]

PR #1131, opened 2026-09-13
fix: only allow alphanumeric OIDC client names to keep environment variables valid

LukasSkywalker:
See #1130 

## Summary by CodeRabbit

- **Bug Fixes**
  - OIDC client creation now validates names consistently.
  - Client names must contain only letters and numbers; hyphens and other special characters are rejected with a clear error message.


cmd/tinyauth/create_oidc_client.go
```diff
@@ -26,10 +26,9 @@ func createOidcClientCmd() *cli.Command {
...
 			clientName := args[0]
 
-			match, err := regexp.MatchString("^[a-zA-Z0-9-]*$", clientName)
-
-			if !match || err != nil {
-				return errors.New("client name can only contain alphanumeric characters and hyphens")
+			// Only alphanumeric since the name will be turned into environment variable names
+			if !regexp.MustCompile(`^[a-zA-Z0-9]+$`).MatchString(clientName) {
+				return errors.New("client name can only contain alphanumeric characters")
 			}
```
LukasSkywalker, line 32: Change the regex match method as well: `MatchString(pattern string, s string)` only returns an error if the pattern parsing failed. With `MustCompile()` and `MatchString(s string)`, this check is moved to compile time.

PR #1154, opened 2026-09-21
refactor: use modular approach for k8s extractors

steveiliop56:
## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes Ingress discovery, including ingress metadata, annotations, namespaces, and hosts.
  * Added wildcard-aware hostname matching for Kubernetes-hosted applications.
  * Kubernetes resource discovery now supports checking resource availability at startup and continues when individual resources are inaccessible.

* **Bug Fixes**
  * Improved domain normalization and case-insensitive application matching.
  * Invalid or non-ASCII domains are now safely excluded from lookups.
  * Added warnings when ingress rules may not cover all request paths.


contre95:
Hey @steveiliop56, When I thought about tinyauth support for k8s gw api I wanted it to  support httproutes only with the `- type: ExternalAuth` and don't have to create a `ReferenceGrant` on the Tinyauth namspace. Is that still the idea?

steveiliop56:
@contre95 not quite. Tinyauth will keep support for the Ingress API for the time being but there will be no watcher for the Gateway API.

Instead, we will use a new CRD designed specifically for Tinyauth ACLs (see https://github.com/tinyauthapp/tinyauth/pull/1155). This CRD will make ACLs in Kubernetes much more easy as it will no longer need to parse the routing definitions themselves and/or play with annotations. The CRD will behave in the same way the Docker labels and the configuration file works.

The new CRD based approach will also pave the way for more Kubernetes-based features such as dynamic OpenID Connect client registration through manifests. 

Keep in mind that nothing will change in the way the Ingress watcher currently works.

